Issue #48, opened 2014-10-01
Preserve bash history


rburny:
Please consider saving Bash history for each pane. This can be done relatively simply (on Bash side) by using following commands:

```
# save history
history -w $HISTORY_FILE
# restore
history -r $HISTORY_FILE
```

How to send these commands to each pane is a trickier question, but I believe tmux's `send-keys` functionality can achieve this - at least when no program is running in given pane.

PR #49, opened 2014-10-17
Bash save and restore history feature

bruno-:
Implementation of the feature discussed in #48.
Thanks @rburny for all the work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - create `expect` script that can fully restore tmux environment
 - fix default save command strategy `ps` command flags. Flags are different for
   FreeBSD.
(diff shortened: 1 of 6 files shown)


foice:
This is a great feature and I think I have used it succesfully for many years now. Still every now and then it fails. I got back to my beloved pane and I find that it contains the history of other bash sessions, maybe other panes ... unclear to me. 
When I save the status I do see the commands 
` history -w '/home/pi/.tmux/my-resurrect///bash_history-0:24.0'` and  `history -r '/home/pi/.tmux/my-resurrect///bash_history-0:24.0'` so I think the `history` command is called correctly. Still the commands in the history after restoring are not what I expect ... do you know of any bug/weak point here?

Issue #57, opened 2014-11-21
Window names are no longer dynamic after restore


frontierpsycho:
I started using tmux-resurrect recently, and I really like it, but I seem to have a problem with window names.

Before I started using tmux-resurrect, window names would be updated based what was running in them at the time. When working on a resurrected session, however, the names of the resurrected windows never change: If I run vim on one of them, the name is still bash. If I quit vim on one that was already called vim (because it was running vim when I saved the session), its name keeps being vim.

rootulp:
Created a [workaround](https://gist.github.com/rootulp/93f67511ae97c20c4d32b600a5ab6a8d) to enable `automatic-rename` on all tmux windows

mayanksuman:
Submitted a pull request #222 that fix this bug.

PR #59, opened 2014-12-08
Implemented test suggested in 

cbertoldi:
https://github.com/tmux-plugins/tmux-resurrect/issues/57 to preserve automatic window renaming.

I've just implemented the suggested check, but I am not saving the window automatic-rename value, because for my taste is good enough ;)

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -78,7 +78,12 @@ new_window() {
 	local window_number="$2"
 	local window_name="$3"
 	local dir="$4"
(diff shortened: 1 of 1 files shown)


pcchou:
:+1:

frontierpsycho:
:+1:

bruno-:
Hi, sorry for not responding to this for a long time.
Is there a reason not to name/rename sessions on restore?

cbertoldi:
No, not really, apart that I was too lazy to implement the whole functionality. The minor change that I did was good enough for my needs.

PR #62, opened 2014-12-13
Save/Restore Buffers Feature

markeissler:
Added support for saving and restoring buffers for panes. Preliminary support is for Bash only (like the recently added Bash history restore support). Also added optional support for ansi color codes in saved pane buffers. Both of these options are disabled by default but can be enabled by adding the following options to the `.tmux.conf` file:

```sh
    set -g @resurrect-save-pane-buffers 'on'
```

```sh
    set -g @resurrect-enable-ansi-buffers 'on'
```

This feature is meant to be used in conjunction with the Bash history save feature but is not dependent upon it. The README has been updated as well.


markeissler:
I'm going to leave this pull request open for now but will likely close it as I work on implementing the auto-save feature which will merge in the work I did here.

markeissler:
I've merged this work (on top of a rebase) with my auto save and restore updates.

PR #67, opened 2015-01-15
Correct ps flag for OpenBSD

SilentGob:
Same flag as FreeBSD is needed


bruno-:
Thanks :+1:

bruno-:
Hi @SilentGob ,
do you know if command `ps -Ao "command"` would on FreeBSD?

According to ps command posix reference [here](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/ps.html#tag_20_96) the `-A` flag should work.
If it does I'm thinking `-A` could be used for all the systems.

bruno-:
I figured I can check this myself.
I fired up a FreeBSD instance on digital ocean and this seems to work okay.

PR #74, opened 2015-02-24
Update README.md

ToruIwashita:
`@resurrect-dir` setting is not work with `~`, and it works by removing quates.


bruno-:
This is now fixed. A path with `~` will work fine. Example: `set -g @resurrect-dir '~/foo'`

PR #79, opened 2015-03-17
Save and restore tmux pane contents

laomaiweng:
This PR adds an option to save the pane contents using tmux's `capture-pane`
command, and restore it upon resurrect using good ol' `cat` (as tmux doesn't
seem to provide a command to set the pane contents).

Having tried this at home, it sure feels a lot more like I never quit tmux
after a reboot rather than with all my panes devoid of any output.

This resembles PR #62 but is apparently a lot less complex (not sure why).

Please comment!

### DIFF
--- a/scripts/helpers.sh
+++ b/scripts/helpers.sh
@@ -51,6 +51,11 @@ remove_first_char() {
 	echo "$1" | cut -c2-
 }
 
+capture_pane_contents_option_on() {
+	local option="$(get_tmux_option "$pane_contents_option" "off")"
+	[ "$option" == "on" ]
+}
+
 save_bash_history_option_on() {
 	local option="$(get_tmux_option "$bash_history_option" "off")"
 	[ "$option" == "on" ]
@@ -79,6 +84,11 @@ resurrect_file_path() {
 
 last_resurrect_file() {
 	echo "$(resurrect_dir)/last"
+}
+
+resurrect_pane_file() {
+	local pane_id="$1"
+	echo "$(resurrect_dir)/pane_contents-${pane_id}"
 }
 
 resurrect_history_file() {
--- a/scripts/save.sh
+++ b/scripts/save.sh
@@ -111,8 +111,8 @@ pane_full_command() {
 
 capture_pane_contents() {
 	local pane_id="$1"
-	local start_line="0"
-	[[ "$(get_tmux_option "$pane_contents_area_option" "visible")" == "full" ]] && start_line="-$2"
+	local start_line="-$2"
+	[[ "$(get_tmux_option "$pane_contents_area_option" "full")" == "visible" ]] && start_line="0"
 	tmux capture-pane -epJ -S "$start_line" -t "$pane_id" > "$(resurrect_pane_file "$pane_id")"
 }
 
--- a/scripts/save.sh
+++ b/scripts/save.sh
@@ -113,7 +113,7 @@ capture_pane_contents() {
 	local pane_id="$1"
 	local start_line="0"
 	[[ "$(get_tmux_option "$pane_contents_area_option" "visible")" == "full" ]] && start_line="-$2"
-	tmux capture-pane -ep -S "$start_line" -t "$pane_id" > "$(resurrect_pane_file "$pane_id")"
+	tmux capture-pane -epJ -S "$start_line" -t "$pane_id" > "$(resurrect_pane_file "$pane_id")"
 }
 
 save_shell_history() {
--- a/scripts/save.sh
+++ b/scripts/save.sh
@@ -78,7 +78,8 @@ state_format() {
 }
 
 dump_panes_raw() {
-	tmux list-panes -a -F "$(pane_format)"
+	local format="${1:-$(pane_format)}"
+	tmux list-panes -a -F "$format"
 }
 
 dump_windows_raw(){
@@ -110,7 +111,9 @@ pane_full_command() {
 
 capture_pane_contents() {
 	local pane_id="$1"
-	tmux capture-pane -ep -t "$pane_id" > "$(resurrect_pane_file "$pane_id")"
+	local start_line="0"
+	[[ "$(get_tmux_option "$pane_contents_area_option" "visible")" == "full" ]] && start_line="-$2"
+	tmux capture-pane -ep -S "$start_line" -t "$pane_id" > "$(resurrect_pane_file "$pane_id")"
 }
 
 save_shell_history() {
@@ -207,9 +210,9 @@ dump_state() {
 }
 
 dump_pane_contents() {
-	dump_panes |
-		while IFS=$d read line_type session_name window_number window_name window_active window_flags pane_index dir pane_active pane_command full_command; do
-			capture_pane_contents "$session_name:$window_number.$pane_index"
+	paste -d"$d" <(dump_panes) <(dump_panes_raw "#{history_size}") |
+		while IFS=$d read line_type session_name window_number window_name window_active window_flags pane_index dir pane_active pane_command full_command history_size; do
+			capture_pane_contents "$session_name:$window_number.$pane_index" "$history_size"
 		done
 }
 
--- a/scripts/save.sh
+++ b/scripts/save.sh
@@ -108,6 +108,11 @@ pane_full_command() {
 	$strategy_file "$pane_pid"
 }
 
+capture_pane_contents() {
+	local pane_id="$1"
+	tmux capture-pane -ep -t "$pane_id" > "$(resurrect_pane_file "$pane_id")"
+}
+
 save_shell_history() {
 	local pane_id="$1"
 	local pane_command="$2"
@@ -201,6 +206,13 @@ dump_state() {
 	tmux display-message -p "$(state_format)"
 }
 
+dump_pane_contents() {
+	dump_panes |
+		while IFS=$d read line_type session_name window_number window_name window_active window_flags pane_index dir pane_active pane_command full_command; do
+			capture_pane_contents "$session_name:$window_number.$pane_index"
+		done
+}
+
 dump_bash_history() {
 	dump_panes |
 		while IFS=$d read line_type session_name window_number window_name window_active window_flags pane_index dir pane_active pane_command full_command; do
@@ -216,6 +228,9 @@ save_all() {
 	dump_windows >> "$resurrect_file_path"
 	dump_state   >> "$resurrect_file_path"
 	ln -fs "$(basename "$resurrect_file_path")" "$(last_resurrect_file)"
+	if capture_pane_contents_option_on; then
+		dump_pane_contents
+	fi
 	if save_bash_history_option_on; then
 		dump_bash_history
 	fi
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -210,10 +210,20 @@ restore_pane_layout_for_each_window() {
 		done
 }
 
+restore_pane_contents() {
+	awk 'BEGIN { FS="\t"; OFS="\t" } /^pane/ { print $2, $3, $7, $10; }' $(last_resurrect_file) |
+		while IFS=$d read session_name window_number pane_index pane_command; do
+			if ! is_pane_registered_as_existing "$session_name" "$window_number" "$pane_index"; then
+				local pane_id="$session_name:$window_number.$pane_index"
+				local read_command=" cat '$(resurrect_pane_file "$pane_id")'"
+				tmux send-keys -t "$pane_id" "$read_command" C-m
+			fi
+		done
+}
 restore_shell_history() {
 	awk 'BEGIN { FS="\t"; OFS="\t" } /^pane/ { print $2, $3, $7, $10; }' $(last_resurrect_file) |
 		while IFS=$d read session_name window_number pane_index pane_command; do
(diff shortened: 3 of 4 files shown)


bruno-:
Hey,
thanks for getting involved and investing the time to write this feature.

I think we definitely want to have this feature.
I didn't review the code, nor tested this but will in the next couple days (so please don't get discouraged until then LOL).

Which shell did you use to test this - bash, zsh?

markeissler:
@laomaiweng Hmm. What's up with the limitation "Currently only the visible area of each pane is saved and restored."? My goal was always to save the complete scrollback buffer automatically and then automatically restore it, the way OSX attempts to resurrect terminal.app sessions after reboot. The buffer feature was supposed to work hand-in-hand with the the excellent @rburny history feature. But I also wanted the process to be 100% transparent without the appearance of any commands sent to the visible terminal. As I continued my journey down the rabbit hole, my code ended up getting more complex to the point that I ended up just forking the project all together into [tmux-resurrect-ng](https://github.com/markeissler/tmux-resurrect-ng) just so I could avoid having to make complex PR requests.

One of the things I'll point out is that @bruno- question about `bash` and `zsh` are important. `zsh` is like the new `csh` these days. My PR supported only `bash` just as a starting point because the original history feature from @rburny was limited to `bash` and I wanted to avoid user confusion.

laomaiweng:
@bruno- This was tested with bash but the code doesn't rely on any particular shell being used, so it should work with any shell. That is, as long as the shell understands `cat <somefile>`, which is needed to restore the pane contents.

@markeissler This limitation was only effective for the first commit of this PR. The second commit adds the ability to save/restore the full pane contents (i.e. including the history).
I still let the default be to only save/restore the visible contents, to make the feature less storage-intensive: 30kB tops for a 81x273 terminal window vs. unbounded size depending on the value of tmux's `history-limit` &ndash; mine is typically 100000, which when filled (and it happens) would weigh in at around, say, 5MB. Currently this is not so much of an issue, since only the latest state has its pane contents and bash history saved, but if tmux-resurrect were at some point to keep those files around with each state, it could become problematic (in particular when combined with tmux-continuum).
What's more, in my workflow the visible pane contents often contain enough context information to get me going after a resurrect.

As it is, the pane contents are saved and restored as is, without any further processing, so the process does have a few visible artifacts when restoring:
* the initial shell prompt (and any output from .bashrc or whatever init script or user activity) is still visible above the restored data (though you have to scroll all the way to the top to see it);
* a `cat` command is visibly issued to restore the data (though, yet again, you have to scroll all the way to the top to see it, and with bash and `HISTCONTROL=ignorespace` set it doesn't show in the bash history);
* if tmux-resurrect doesn't restore any running process, a shell prompt (the "current" prompt) is issued after the restored pane contents, which sort of comes as an "additional" prompt if the last line of the restored pane contents was itself a prompt;
* if tmux-resurrect does restore a running process (e.g. vim), it's output will appear twice: first in the restored pane contents (the process was running during the save) then upon resurrect of the process.

I must say that for the time being, none of these glitches came across as annoying to me. I didn't think they'd be seen as harmful considering that the bash history save/restore feature already adds some similar minor glitches. I'm sure some of these could be fixed, but that'd add significant complexity to a rather simple PR for a minor gain IMO.
Also note that saving the pane contents is completely transparent thanks to tmux's `capture-pane` command.

bruno-:
Hi @laomaiweng,
thanks for pointing out all the glitches. I think all of them are ok as long as we mention them transparently to the users in the readme.

Quick question, did you try restoring data in the pane with something like this: `tmux new-window "cat some_file; bash"`.
It seems that one has the benefit of not showing the first prompt and the initial `cat` command. I tried it just after reading your post and it worked, but I might be missing something. In fact, that solution definitely misses proper shell detection (we don't want to just hardcode `bash`).

I'm gonna try running this pull request.

scripts/helpers.sh
```diff
@@ -81,6 +86,11 @@ last_resurrect_file() {
 	echo "$(resurrect_dir)/last"
 }
 
+resurrect_pane_file() {
+	local pane_id="$1"
+	echo "$(resurrect_dir)/pane_contents-${pane_id}"
```
bruno-: I can see myself having hundred thousand `pane_contents` files in a matter of days (with tmux-continuum saving environment every 5 minutes).
Ideally we'd wanna have a new directory with a save timestamp within `resurrect_dir` for each save.

I not sure if that is easily implemented, so I'm just starting the discussion..

scripts/save.sh
```diff
@@ -108,6 +109,13 @@ pane_full_command() {
 	$strategy_file "$pane_pid"
 }
 
+capture_pane_contents() {
+	local pane_id="$1"
+	local start_line="0"
+	[[ "$(get_tmux_option "$pane_contents_area_option" "visible")" == "full" ]] && start_line="-$2"
+	tmux capture-pane -ep -S "$start_line" -t "$pane_id" > "$(resurrect_pane_file "$pane_id")"
```
bruno-: I was just reading up about `capture-pane` flags, what do you think about adding `-J` option?
So, after restore, if a user resizes pane/window, the text should automatically re-wrap to the new width.

laomaiweng:
Hi @bruno-,
I didn't try anything like `tmux new-window "cat some_file; bash"`. But apart from the missing shell detection, this one is a no-goer for me: my .bashrc spews out a lot of output, and I don't want this output to happen *after* the restored pane contents.
On a side note, I just tried `tmux run-shell -t "$pane_id" "cat some_file"`, which doesn't work to restore the pane contents.

scripts/helpers.sh
laomaiweng: Much like `bash_history` files, those files don't add up with saves.
Though there is indeed some clutter from previous saves if some panes where destroyed (e.g. if a previous save layout had panes up to 3.4 in window 3 and the current one only has up to 3.2, there will be stale `pane_contents-session:3.3` and `pane_contents-session:3.4` files in the `.tmux/resurrect` directory), `pane_contents` files are replaced with each save, so their numbers currently doesn't grow (except if you create more panes).

However as I mentioned earlier, if tmux-resurrect were to keep those files around attached to each save, then the growth could become problematic. I must admit I have an upcoming PR in my todo-list to create a tarball with the additional files for each save. A quick test tarball of my `pane_contents` files shows a compression ratio of 8 for gzip, 16 for bzip2 and 21 for xz. That ought to make the total size much more manageable, while reducing clutter (1 `tmux_resurrect` + 1 tarball per save).
But as said in the previous paragraph that is currently a problem for another time.

scripts/save.sh
laomaiweng: I hadn't even noticed this option as it doesn't appear in the options summary in my tmux manual page (it's only mentioned in the text description for `capture-pane`).
It does look like a good addition indeed. I just gave it a shot and it seems to be working, so I'll commit it.

bruno-:
Hi @laomaiweng,
another approach for not showing `cat file` at the top of the pane could be:
- get pane tty with `tmux display -F "#{pane_tty}"`. This should return something like `/dev/ttys123`
- just append text to the returned file `cat some_file >> /dev/ttys123` and it should appear in the pane

The downside to this seems to be that it does not show command prompt after the text is appended to the pane (the user has to press enter). I just tried this so I'm not sure if there are other downsides and how to handle them.

bruno-:
@laomaiweng, I'm testing the feature and it feels great! Thanks for the work on this so far.

I do have a couple observations:
- when a pane is empty (no history other then the single shell prompt line) we're both saving and restoring pane history. If possible, we should probably prevent this.
- for panes with a shell prompt at the top of the screen, the \*whole\* screen is saved - even the completely empty lines before the shell prompt. When restoring, this has the unwanted effect of outputting those empty lines and leaving the restored pane shell prompt at the bottom of the screen.
I think the solution for this is to trim the empty lines from the bottom of the `pane_contents*` files. This isn't hard, I've worked on this same problem in tmux-logging plugin, you can see the solution I came up with [here](https://github.com/tmux-plugins/tmux-logging/commit/386eec0bc6b3a6a718d12647a4c688db98db6b58#diff-7d143e6f5e114cff05b0bc1b8953ca4bR37).
- space consumption: full pane contents area files are only ~15MB for an environment of 100+ panes I have (all the files combined). I have a history limit of 50k, but many panes don't reach it. I propose we make the 'full' pane contants area the default because it is so much nicer. This will be even less relevant as we work on the zipping these files.

bruno-:
@laomaiweng, the feedback above may be overwhelming .. what do you think about this workflow moving forward:
- I've merged your commits to [resurrect-pane-contents](https://github.com/tmux-plugins/tmux-resurrect/tree/resurrect-pane-contents) branch. Now, we can both submit commits for this feature to that branch. 
- since you "own" this feature, I'll also be opening pull requests for additions to that branch and ask you to review.
- when we're happy with how things work, we'll merge that branch into master

What do you say?

laomaiweng:
Hi @bruno-,
Using `#{pane_tty}` sounds great! It would completely bypass the shell, which is neat.
The missing command prompt at the end can be solved in 2 ways:
* `cat "$pane_file" >> "$pane_tty" ; tmux send-keys -t "$pane_id" C-m` => additional prompt if no process to resurrect and the last line of the restored pane contents was a prompt
* `head -c -1 "$pane_file" >> "$pane_tty"` => strips the terminating newline, so the cursor appears to be located at the end of the last prompt (if any) in the restored pane contents

So to wrap up, we could go about it this way:
* pane was running a shell (it is my understanding that this is true when `pane_full_command` from the resurrect file is ":", is that correct?) => `head -c -1 ...`
* pane was running a process that we are *not* going to resurrect => `cat ... ; tmux send-keys ...`
* pane was running a process that we *are* going to resurrect => `cat ... ; tmux send-keys ...` (since we're going to use `tmux send-keys` to resurrect the process afterwards, it makes more sense to display a prompt before process resurrection occurs)

Which boils down to:
* pane was running a shell => `head -c -1 ...`
* pane was not running a shell => `cat ... ; tmux send-keys ...`

In the end this should solve glitches 2 and 3 (see my above comment, I numbered the glitches) while not adding significant complexity.

*However*, there is an unexpected difficulty: pane creation is done asynchronously, and under some circumstances there's a race condition between shell startup and contents restoration.
E.g., if the shell is a little slow to initialize (as is the case on my machine, since my bashrc does quite a lot of things), it may happen that tmux-resurrect tries to cat the pane contents before the prompt is displayed, resulting in pane contents restoration to happen before display of the first prompt (and whatever output from the bashrc).
This doesn't happen when sending the command with `tmux send-keys` (without `> "$pane_tty"`, i.e. as things are currently in the PR) since the command is then processed by the shell and hence doesn't get run before the shell is ready.
I'm not quite sure how to properly solve this yet.

What's more, this `> "$pane_tty"` business introduces a new glitch (would be glitch 5): if the shell is waiting but not at a primary prompt (e.g. a secondary prompt, or within a shell builtin such as `read`...), the output is restored and the cursor correctly positioned, but the secondary prompt/shell builtin is no longer active. Thus any further input will be used as if typed at a primary prompt, which would be a little confusing to the user.
This doesn't apply to the current way of restoring pane contents, since there's always a new prompt at the end (glitch 3).

All this makes me wonder if things aren't better off as they are now. Any thoughts?

laomaiweng:
W.r.t. your observations:
* We can check whether the saved pane contents is only 1 line, and not restore it in this case (this no longer works if bashrc outputs stuff before the 1st prompt, but that's okay, it doesn't break anything).
* Indeed, I just noticed this while testing the `pane_tty` approach earlier today. However your solution strips possibly legitimate blank lines (I admit that's a corner case). I think we can fix this properly using `#{cursor_y}` and `capture-pane -E`.
* Agreed to make `full` pane contents saving the default!

I got a little afraid that copy-mode would break pane capture, but it appears it doesn't: when a pane is in copy-mode, its contents are captured as if copy-mode wasn't enabled.
However we might need to look out for panes using the alternate screen. I have this feature disabled as I don't like it, so I have not tested it, but I think this will need some looking into. `capture-pane` has a `-a` option to capture the alternate screen.

laomaiweng:
@bruno-, I'm perfectly happy with the abundant feedback!

I'm fine with the workflow you propose. However feel free to bypass me if you will, when committing to the branch: I don't have a very strong sense of property towards the code I contribute. :)

laomaiweng:
Hi @bruno-,

I am currently testing a solution to the saving and restoration of trailing blank lines, that does not involve blindly trimming those lines (as they may be legitimates lines of program output).

It makes use of tmux's `#{cursor_y}` and `#{keypad_flag}` variables. Will push once I'm satisfied it works. If you want to test it on your side without getting it into the branch (I consider it quite experimental), I can also provide a patch.

bruno-:
Hi @laomaiweng,
once again, thank you for writing this.
At first I wanted to get this branch "perfect" before merge, but given my schedule, that's probably not possible. So the branch is merged now.
I'll mention this feature in the readme and mark it experimental (for now).

Feel free to open a pull request for removing the empty lines. Hopefully the solution with tmux's `#{cursor_y}` is not too complex? I doubt anyone will ever complain about legit empty lines at the bottom being removed..

bruno-:
We've made some progress with this feature, you can check out issue #81. The last "serious" remaining todo is implementing compression of pane content files.

If anyone is using this please update the plugin and let me know if things work for you. Other suggestions are welcome.

Issue #81, opened 2015-04-02
Todos for improving save and restore tmux pane contents feature


bruno-:
Open todos for "save and restore tmux pane contents" feature which was contributed by @laomaiweng in #79.
The todos were discussed in that PR, but now that that's merged, the todos are extracted here so we don't forget them.

- [ ] pane content files are stored in `~/.tmux/resurrect/` dir. Maybe store them in `~/.tmux/resurrect/pane_contents/` so they don't clutter main dir?
- [ ] implement compression of pane content files `gzip` or `bz2`. Idea by @laomaiweng, brought up [here](https://github.com/tmux-plugins/tmux-resurrect/pull/79#discussion_r26645399)
- [ ] currently pane contents are displayed with a simple `cat` command which is left displayed at the top of the pane. Ideas for doing this unobtrusively (all of these have downsides):
   - `tmux new-window "cat pane_contents_file; bash"` - will restore pane contents and start a shell in the pane. [Idea here](https://github.com/tmux-plugins/tmux-resurrect/pull/79#issuecomment-82662701).
  - `cat some_file >> /dev/ttys123` where tty is tty of current pane. [Idea brought up here](https://github.com/tmux-plugins/tmux-resurrect/pull/79#issuecomment-82956797).
- [ ] do not output trailing empty lines when pane content is restored. @laomaiweng has an interesting idea brought up [here](https://github.com/tmux-plugins/tmux-resurrect/pull/79#issuecomment-88101600). The alternative is to implement something like [this](https://github.com/tmux-plugins/tmux-logging/commit/386eec0bc6b3a6a718d12647a4c688db98db6b58#diff-7d143e6f5e114cff05b0bc1b8953ca4bR37).
- [ ] investigate capturing panes with alternate screen. Brought up [here](https://github.com/tmux-plugins/tmux-resurrect/pull/79#issuecomment-83170691).
- [ ] do not save pane content history when a pane is empty (no command was ever executed inside it)

absorber:
Perhaps also consider replacing `/home/user` to `$HOME` in `~/.tmux/resurrect/last`, so that resurrect can work in situations where the saving user is different from the restoring user (on different machines for example).

As can be observed from this bug - https://github.com/tmux-plugins/tmux-resurrect/issues/77

22459:
[...]

PR #82, opened 2015-04-10
restore pane contents directly to TTY, not via cat

sunaku:
This commit restores pane contents directly to the pane TTY instead of
executing a "cat" command inside that pane's attached shell session.
Therefore no extraneous "cat" commands are added to your shell history.

It also folds blank lines and omits the final newline during restoration.

https://github.com/tmux-plugins/tmux-resurrect/pull/79#issuecomment-82956797

https://github.com/tmux-plugins/tmux-resurrect/issues/81

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -216,7 +216,7 @@ restore_pane_contents() {
 			if ! is_pane_registered_as_existing "$session_name" "$window_number" "$pane_index"; then
 				local pane_id="$session_name:$window_number.$pane_index"
 				local pane_tty=$(tmux display-message -p -F "#{pane_tty}" -t "$pane_id")
-				cat -s "$(resurrect_pane_file "$pane_id")" | perl -pe "chomp if eof" > "$pane_tty"
+				cat -s "$(resurrect_pane_file "$pane_id")" > "$pane_tty"
 			fi
 		done
 }
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -215,8 +215,8 @@ restore_pane_contents() {
 		while IFS=$d read session_name window_number pane_index pane_command; do
 			if ! is_pane_registered_as_existing "$session_name" "$window_number" "$pane_index"; then
 				local pane_id="$session_name:$window_number.$pane_index"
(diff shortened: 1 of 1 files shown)


sunaku:
Hey @bruno- Any chance of merging this? :sweat_smile: It greatly improves `restore_pane_contents()`. :rocket:

bruno-:
Hey @sunaku,
sorry for waiting on the response..
I just looked into the pull request: is there a way to remove the dependency  on `perl`?

bruno-:
Also, I know the solution proposed in [this comment was discarded](https://github.com/tmux-plugins/tmux-resurrect/pull/79#issuecomment-82662701), but frankly I think it's the most elegant. How about re-evaluating that one?

Also, implementing it should be really simple..

sunaku:
Thanks for your response! :bell:  Perl dependency is now removed. :scream: 

I looked at your proposal: you can detect the current shell by looking at `$SHELL` or better yet `$0`.  But we'd also need to restore the window ourselves (using `tmux new-window`) rather than catting into an already restored window. :construction:

bruno-:
Hey,
thanks for being so responsive on this.
So last time I tried `cat`ing to `#{pane_tty}` there was the downside of command prompt not being visible after the action...

Anyway, I just played with implementing [this idea](https://github.com/tmux-plugins/tmux-resurrect/pull/79#issuecomment-82662701) again. It seems the cleanest approach and we're working solely with tmux api, eg no outside files etc (which is good). I pushed the implementation to master.

@sunaku can you please try pulling down latest plugin version and check how things work for you?
Btw. sorry for not going with implementation in this pull request.. I hope you'll like this other solution even better.

bruno-:
If anyone is interested in implementation, here's the commit https://github.com/tmux-plugins/tmux-resurrect/commit/c85a3b4487b2d97c3bfc92c583c595c5fc83603d

sunaku:
Hey @bruno- your new implementation works great!  :+1: This pull request is no longer needed. :fire:

bruno-:
Great @sunaku , thanks for checking this out!

PR #88, opened 2015-06-10
try to treat inline strategy as alias for strategy

dmotylev:
if strategy with that name exists apply it otherwise use it as inline
strategy. This makes vim restore strategy usable with MacVim. 

    set -g @resurrect-strategy-vim 'session'
    set -g @resurrect-processes 'vim man less tail "~Vim->vim"'


bruno-:
Hey, thanks, this makes sense!

This is merged via the command line (I just wanted to add a comment). You're still the committer.

PR #97, opened 2015-08-19
keep the window names dynamic after restore

knight42:
I append an option named `rename_option` after `window_name` to store the value of `automatic-rename` for each pane.

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -23,7 +23,8 @@ RESTORE_PANE_CONTENTS="false"
 is_line_type() {
 	local line_type="$1"
 	local line="$2"
-	grep -q "^$line_type" <<< "$line"
+	echo "$line" |
+		\grep -q "^$line_type"
 }
 
 check_saved_session_exists() {
@@ -38,8 +39,8 @@ pane_exists() {
 	local session_name="$1"
 	local window_number="$2"
 	local pane_index="$3"
-	tmux list-panes -t "${session_name}:${window_number}" -F "#{pane_index}" 2>/dev/null | \
-		grep -q "^$pane_index$"
+	tmux list-panes -t "${session_name}:${window_number}" -F "#{pane_index}" 2>/dev/null |
+		\grep -q "^$pane_index$"
 }
 
 register_existing_pane() {
@@ -78,8 +79,8 @@ is_restoring_pane_contents() {
 window_exists() {
 	local session_name="$1"
 	local window_number="$2"
-	tmux list-windows -t "$session_name" -F "#{window_index}" 2>/dev/null | \
-		grep -q "^$window_number$"
+	tmux list-windows -t "$session_name" -F "#{window_index}" 2>/dev/null |
+		\grep -q "^$window_number$"
 }
 
 session_exists() {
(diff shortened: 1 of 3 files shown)


knight42:
@bruno- errrr... could you update the test files? I think it should work without breaking anything

bruno-:
Hey,
thanks for working on this PR, but I don't think this is ready to be merged. The biggest issue I see is that it changes format of `tmux-resurrect` save file in a way that's not backwards compatible.. and actually tests fail because of that.
Is there a way to make this a backwards compatible change?

Also, the PR contains a couple other changes I'm not sure we want:

- vim modeline strings - I'm a vim user too, but we probably don't want anything editor specific in the source code
- changes related to `join_str` function - this is mostly a stylistic change no? Let's focus on the core problem for now.
- change from `\grep` to just `grep`: actually I think we want to have `\grep`, it's related to this [bash feature](http://stackoverflow.com/questions/15691977/why-start-a-shell-command-with-a-backslash). There was a user that had issues with `grep` in `tmux-copycat` plugin because he had it wrapped in a function... so we better leave this as-is or handle separately.
- there some inconsistent indentation brought in with these changes..

Please don't get discouraged with the objections.. we DO want to handle this issue..

knight42:
@bruno- 
Thanks for your prompt response! I have reverted my changes.

Firstly, I am sorry for my ignorance of the usage of `\grep`. I misunderstood that it is a typo so I fixed it... Perhaps adding a comment to explain it would be better.

Then I want to talk about my new implementation. In brief, I want to REUSE the var `window_active`.
Since the value of `window_active` is either 0 or 1, it occured to me that we can reuse it to store the value of `automaic-rename`. Here is how it works.

Let's set `RENAME_FLAG` to be `2`, if a window's name is dynamic, then execute:
`window_active=$((window_active | $RENAME_FLAG))`
and save the value of `window_active`.
When restoring, if `$((window_active & $RENAME_FLAG)) -ne 0`, which means the window name is dynamic, then there is no need for us to specify the window's name.

I don't know whether I have missed sth, I think this implementation is backwards compatible(some changes need to be made to get the real value of `window_active` though), but it still needs to modify the save file. Thus it may fail the CI test :-( . 
However, I consider it unavoidable to modify the save file if we want to add this feature. The matter is how greatly it will be changed. If you accpet my implementation, could you please update your test files according to my idea? Thank you :-D

Lastly, let's talk about sth relatively unimportant.
As for the `join_str` function, yeah it is totally a stylistic change. I admit that I am kinda too agressive, but I think the strings can be joined in an a bit more elegant way, so I want to improve it. If you object the change, it is ok, please let me know. After all, it is just a personal issue.

When speaking of the conflict between spaces and tabs,  it is one of the eternal HolyWars among programmers ;-) , right? I know it may be somewhat unreasonable to you, but I recommend that we replace tabs with spaces because tabs are "ambiguous", which can be expanded to 2, 4 and even 8 spaces in different editors, IDEs, but with spaces, everything is obvious.
What's more, tabs in files can be expanded to spaces conveniently in Vim, just set `tabstop` to anything you like(but I recommend 4...) and execute `:retab`. All done!

Since I am not a native English speaker, I hope I have made my points clear. If not, feel free to ask. As you said, we DO want to handle this issue.

Cheers,
Knight

knight42:
As expected, it fails :-( Anyway, I have to go to sleep now...

scripts/restore.sh
```diff
@@ -152,7 +182,12 @@ new_pane() {
 	local window_name="$3"
 	local dir="$4"
 	local pane_index="$5"
+	local rename_option="$6"
 	local pane_id="${session_name}:${window_number}.${pane_index}"
+
+	# The following line may be executed repeatedly when creating multiple panes.
+	[[ $rename_option -eq 0 ]] && tmux rename-window -t "$session_name:$window_number" "$window_name"
```
knight42: rename first window's name to the correct name

Issue #103, opened 2015-09-16
Missing previous directory after doing restore


linpc:
Hi,

Found that when restoring tmux session, the previous working directory won't be recovered. This causes commands like `vim`, `less` failed to find the correct file in the previous directory.

I will fire a PR try to fix this.

==
OS: Mac OS X
tmux version: 2.0
`~/.tmux/resurrect/last`: https://gist.github.com/linpc/0b5f7f7ba41274f690a2
tmux.conf setting:
```conf
run-shell ~/src/tmux-resurrect/resurrect.tmux
set -g @resurrect-processes 'ssh telnet ping "git log"'
```

bruno-:
Hey,
this is one of the core plugin features and should be working.

Previously we had similar bug reports in #80 and #50.

Apparently `last-working-dir` zsh plugin was breaking this. If you have it installed, can you disable it and try again to see if things will work?

Also, can you please try doing this:

1. Try `tmux new-window -c '/tmp'`. If this *does not* open a new window in `/tmp`, something is wrong, see next step.
2. Try `tmux new-window -c '/tmp' 'pwd; sleep 1000'`. It should show `/tmp` directory on the screen. If yes, then tmux is ok and a user config (eg bashrc, zsh plugins, something else) is wrong.

What do you get when you execute steps 1 and 2?

selurvedu:
This looks similar to #90 as well.

linpc:
Hi @bruno- ,

Thanks for explaining.

I found there's a "cd $HOME" at the 1st line in my `.bashrc` config, and this breaks the `tmux new-window -c $dir` command. (looks like tmux use the parameter before invoking the shell setting?)

bruno-:
Hey,
ok, I think this explains it.

Yep, `.bashrc` (invoked by `bash` not `tmux`) is invoked way later than `tmux new-window -c $dir`.

alanlal:
> Hi @bruno- ,
> 
> Thanks for explaining.
> 
> I found there's a "cd $HOME" at the 1st line in my `.bashrc` config, and this breaks the `tmux new-window -c $dir` command. (looks like tmux use the parameter before invoking the shell setting?)

I had the exact same issue with `tmux-resurrect`. It was due to my setting `cd ~` in my `.zshrc` file. I never thought of this. thanks for the idea.

PR #104, opened 2015-09-16
Recover the working directory for command to run well

linpc:
- Command like `vim', `less`, or `git log' may need to run in specific
  directory for process could find related files.
- Fix issue #103


bruno-:
Hi @linpc,
I think we found the core issue for this in #103. Based on that I think it's safe to close this PR.

PR #125, opened 2016-01-29
Fix comparision operators. Fixes #124

### LINKED ISSUE #124: errors in save.sh and restore.sh
You are using a single `=` which is an assignment and not a comparison operator.

restore.sh https://github.com/tmux-plugins/tmux-resurrect/blob/master/scripts/restore.sh#L277
should be:
`if [ "$pane_command" == "bash" ]; then`

save.sh https://github.com/tmux-plugins/tmux-resurrect/blob/master/scripts/save.sh#L149
should be:
`if [ "$pane_command" == "bash" ] && [ "$full_command" == ":" ]; then`

HeathNaylor:
Fixes #124


bruno-:
Thank you, I guess we should be consistent with the equality.

Just in case someone wonders how this "noob" code got into the repo in the first place http://stackoverflow.com/questions/3265803/bash-string-equality

PR #126, opened 2016-02-01
fix soft link error in mac and set pane automatic-rename after restore

hiberabyss:
1. Save a session in osx, after restarting, the `last` symbol file will link to incorrect file whose name end with `%Y-%m-%dT%H_%M_%S`.
1. Set pane `automatic-rename` after restoring.

### DIFF
--- a/scripts/helpers.sh
+++ b/scripts/helpers.sh
@@ -98,7 +98,7 @@ resurrect_dir() {
 }
 
 resurrect_file_path() {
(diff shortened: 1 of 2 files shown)


bruno-:
Hey,
I've never had this issue with `last` linking to incorrect file name... colon is a valid filename character (although unusual yes).

About the `automatic-rename`: unfortunately we can't force that on the users.

PR #133, opened 2016-03-02
Prevent history command from being recorded

Purg:
If the assumed HISTCONTROL settings are not present, the history command is written to the saved file, causing the history command to reappear when restoring panes. This longer command prevents the saving of the command regardless of the HISTCONTROL settings by writing it out and then copping off the final history line containing the write command. The long command also ends with clearing and restoring the local history with what was recorded in order to prevent the history writing command from leaking into the saved file when performing multiple tmux save actions.


bruno-:
Hey,
I'd like to accept this pull request if it's an improvement.

Unfortunately the above problem explanation is so dense I'm unable to figure out what's wrong. Would you mind passing a couple examples?

Purg:
Sorry for the late response.

Basically what's happening for me is that when saving the session the history dump command is being saved with a panes history. This was previously addressed by assuming a specific HISTCONTROL setup. When HISTCONTROL was not setup in the assumed way, the history dump commands were saved in this history dump, causing them to be the most recent command when restoring pane histories.

This change constructs a longer command, but should generically fix the issue where the history dump command is saved. With this commit, after a tmux restore, the last command in the history of a pane is the last command actually performed by the user, not the history dump command.

Purg:
For example, before this change, the last command in the history of all restored panes was the `` history -w '$(resurrect_history_file "$pane_id")'`` command.

Purg:
Before, the command was assuming ``ignorespace`` was included in the HISTCONTROL env variable, which wasn't the case for me (and probably shouldn't be assumed).

laomaiweng:
I feel the proposed solution is a bit too heavy-handed, as it involves several operations on potentially large history files.

Here is a proposal that works regardless of the value for `HISTCONTROL`:
````history -d "$(history 1 | sed -e 's/^ *\([0-9]*\).*/\1/')" ; history -w <pane_history_file>````
What it does is basically erase itself from history, then write the history file. This works since lines are appended to history before being executed.

Note that there SHOULDN'T be a space at the start of the line (contrary to how things are now in `save.sh`) so that the line DOES get into the history and gets removed. If there was a space at the start of the line, with `HISTCONTROL=ignorespace` or `ignoreboth`, the line wouldn't get into the history at all and we would erase the last legitimate line.

Purg:
Sounds better than what came up with. Commit in a minute.

Purg:
Bump?

PR #149, opened 2016-05-28
Fix session restore failed when command with absolute path

allex:
Fix session restore failed when command with absolute path

### DIFF
--- a/scripts/process_restore_helpers.sh
+++ b/scripts/process_restore_helpers.sh
@@ -94,6 +94,9 @@ _proc_matches_full_command() {
 			return 0
 		fi
 	else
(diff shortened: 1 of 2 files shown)


bruno-:
Hm.. I've never had an issue with this. Can you explain what's wrong with this?

allex:
Not sure for details about command with full path.
i've just found there are some absolute path in session file, just like below:

```
pane	0	3	:test-session	1	:*	1	:/Users/allex/dev/web-tools/kitty	1	vim	:/usr/bin/vim bin/checkout.sh
```
in that case, this panel process will not be restored success (because of the `/usr/bin/vim`)

PR #151, opened 2016-06-29
auto remove old backup files

### LINKED ISSUE #102: Old resurrect-files are not deleted
Old resurrect-files are not deleted, I have tons of them (from today and yesterday 123 files alone). It's only ca 9MB in files after a month of use but still.
Maybe do a resurrect and after that delete all files that are related to the resurrect.

xu-cheng:
remove backup files older than 30 days, but keep at least 5 copies of backup.

Fixes #102.

### DIFF
--- a/scripts/save.sh
+++ b/scripts/save.sh
@@ -252,6 +252,13 @@ dump_bash_history() {
 		done
 }
 
+remove_old_backups() {
+	# remove backup files older than 30 days, but keep at least 5 copies of backup.
+	local -a files
+	files=($(ls -t $(resurrect_dir)/*.txt | tail -n +6))
+	[[ ${#files[@]} -eq 0 ]] || find "${files[@]}" -type f -mtime +30 -delete
+}
+
 save_all() {
 	local resurrect_file_path="$(resurrect_file_path)"
 	local last_resurrect_file="$(last_resurrect_file)"
@@ -274,6 +281,7 @@ save_all() {
 	if save_bash_history_option_on; then
 		dump_bash_history
 	fi
+	remove_old_backups
 }
 
 show_output() {


xu-cheng:
Ping @bruno-

reenberg:
I would also wary much like this to be merged in.

Kriechi:
Does `tail -n +5` work on OSX too? This usually behaves a bit different due to BSD/GNU weirdness.

xu-cheng:
> Does tail -n +5 work on OSX too? 

Yes.

Updated a bit to handle when there is less 5 files. Any chance to review this and maybe merge it? Thanks.

czchen:
Hi @bruno- ,

Any update for this PR?

scripts/save.sh
```diff
@@ -252,6 +252,15 @@ dump_bash_history() {
 		done
 }
 
+remove_old_backups() {
+	# remove backup files older than 30 days, but keep at least 5 copies of backup.
+	local -a files
+	pushd "$(resurrect_dir)" >/dev/null
+	files=($(ls -t ./*.txt | tail -n +6))
```
bruno-: Is there any benefit of using `pushd` + `popd` over just `ls -t $(resurrect_dir)/*.txt`

xu-cheng: Updated to use `ls -t $(resurrect_dir)/*.txt`

bruno-:
Hi,
thank you for the contribution. This is merged locally so closing this PR.

When I tested this on a mac I found the `find -delete` option doesn't work. The problem symptoms are described [here](https://lists.freebsd.org/pipermail/freebsd-hackers/2015-March/047170.html).
To get around this I changed the delete command to use `find ... | xargs rm`.

PR #152, opened 2016-07-05
Fallback for getting the pane_current_path

### LINKED ISSUE #131: Doesn't work on Windpws 10
tmux-resurrect has very limited usefulness on Windows 10.

When I restore the session it will bring back my windows (with names) and panes where I left them. However, it does not remember what path each of these windows were in. This dramatically limits the usefulness of tmux-resurrect for windows users.

timfel:
This might fix #131 and maybe other people's issues where pane_current_path is sometimes not set correctly (happens with Tmux 2.0 on CentOS).

I realize this looks horrendous, but I couldn't find a better solution right now.

### DIFF
--- a/scripts/save.sh
+++ b/scripts/save.sh
@@ -41,7 +41,7 @@ pane_format() {
 	format+="${delimiter}"
 	format+="#{pane_index}"
 	format+="${delimiter}"
-	format+=":#{pane_current_path}"
+	format+=":#{?pane_current_path,#{pane_current_path},#(lsof -t #{pane_tty} | head -1 | xargs -I % readlink -e /proc/%/cwd)}"
 	format+="${delimiter}"
 	format+="#{pane_active}"
 	format+="${delimiter}"


soda0289:
I'm on Bash for Windows 10 and this does work sometimes. Other times I get this saved in tmux_resurrect file:

`pane    0       0       :bash   1       :*      0       :<'lsof -t /dev/pts/0 | head -1 | xargs -I % readlink -e /proc/%/cwd' not ready>        1       bash    :`

Which causes tmux to crash on restore.

soda0289:
It seems if I save the first time it produces the invalid result, but on the second save I get the expected result.

timfel:
Interesting, I might not have noticed because I use it with tmux-continuum with an auto-save interval of 4 minutes. I'm not sure if there is a way to tell tmux to wait for the execution

soda0289:
I will investigate further...

Other people have this issue as well with tmux variables:
http://unix.stackexchange.com/questions/253267/tmux-displays-xzy-not-ready-in-the-status-bar

I think it is this function in tmux that is causing the problem:
https://github.com/tmux/tmux/blob/master/format.c#L240

Will read through the code and try to find possible workaround.

PR #163, opened 2016-10-25
added instructions for restoring to an earlier saved environment

houghjosh:
I just copied over the instructions from [https://github.com/tmux-plugins/tmux-continuum/blob/master/docs/faq.md](url) for restoring an environment that was saved at a previous point in time. I had to hunt around a bit to find the instructions after accidentally saving an empty environment instead of restoring, so I thought having them here could be useful for others in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,16 @@ You should now be able to use the plugin.
 
 - [restoring bash history](docs/restoring_bash_history.md)
 
+**Restoring a previously saved environment**
+
+Here are the steps to restore to a previous point in time:
+
+- make sure you start this with a "fresh" tmux instance
+- `$ cd ~/.tmux/resurrect/`
+- locate the save file you'd like to use for restore (file names have a timestamp)
+- symlink the `last` file to the desired save file: `$ ln -sf <file_name> last`
+- do a restore with `tmux-resurrect` key: `prefix + Ctrl-r`  
+
 ### Other goodies
 
 - [tmux-copycat](https://github.com/tmux-plugins/tmux-copycat) - a plugin for


falzard:
should be added

bruno-:
Hi,
agreed this would be helpful to have. Can you please change the PR so it's just a link to the relevant section in tmux-continuum docs?

z-huabao:
now `prefix` `Ctrl-r` restore default resurrect file `last`.

Can you make `prefix` `Ctrl-r` show a list of resurrect files, we can select the file by `<Up><Down>` or `j` `k`, and last save file on first as default, then we please `<Enter>` to select the file.

ipatch:
link posted above by @houghjosh is now defunct ie. broken. 👇

```
https://github.com/tmux-plugins/tmux-continuum/blob/master/docs/faq.md
```

that said, it'd be nice to have **FAQ** shomewhere in the README for troubleshooting some basic functionality of this plugin.

1. For my use case, I could not restore from my previous tmux session after my box had a kernel panic, and it ended up being due to a permissions issue, **my fault** 🙋‍♂️, that said, a little one or two liner explaining how tmux-resurrect creates a symlink to a `txt` file containing session data within a specified directory would be useful when one wants to restore from an older session state.

Maybe I'm blind, but I did not see anywhere in the README explaining the symlink and how relinking **last** to previous `txt` within the **session** dir could aid in restoring a previous session state.

2. Another use case that has happened to me in the past is opening tmux and mashing my key-binding for save instead of restore, and not having an idea that the previous session states were still on my drive and all i had to do was relink the **last** file to the appropriate `txt` file within the session dir.

> leaving this here for future me 🚀👴🏻

PR #175, opened 2017-01-02
Parse alias when (n)vim sessions are resurrected

izumin5210:
Enable to restore (n)vim sessions when an alias is set to (n)vim command, such as followings

```
alias vim="nvim -c Obsession"
```


bruno-:
Hi, thank you for the contribution. However, I don't think we'd want to accept this PR.

PR #180, opened 2017-02-12
Send keys to the correct pane

mk12:
The process restoring feature of tmux-resurrect did not work at all for me because all the keys got sent to the same pane. I have fixed the `send-keys` commands in `process_restore_helpers.sh` to include the target pane ID, as is already done in `restore.sh`.


bruno-:
Hey, thanks for this one. I added similar changes earlier and now used your PR to add this last change 1a3a094265a5a1173a703005990681236cbdbc4a

PR #191, opened 2017-05-05
remove terminal clearing since it collides with capture-pane-contents

Mellbourn:
The clearing of the terminal after history restore is not a feature. 
It conflicts with `resurrect-capture-pane-contents`. 
If you have both history restoration and content capture turned on, the screen will be restored then promptly erased.


Mellbourn:
closing this in favor of a better pull request based on a branch

Issue #141, opened 2016-04-17
Error when restoring pane content


Limeth:
When I restore the session, an error message appears at the top of all the panes which says a file or directory could not be found.

This is what I have in .tmux.conf:

```
set -g @resurrect-strategy-vim 'session'  # save vim sessions (may require the tpope/vim-obsession plugin)
set -g @resurrect-strategy-nvim 'session'  # save neovim sessions (may require the tpope/vim-obsession plugin)
set -g @resurrect-capture-pane-contents 'on'  # save pane contents
set -g @continuum-restore 'on'  # restore session on tmux start
set -g @continuum-save-interval '5'  # save session every 5 minutes (default is 15)
```

![Screenshot](http://storage5.static.itmages.com/i/16/0417/h_1460899144_2416053_7979c6e2d6.png)

bruno-:
This seems to be the problem related to `@resurrect-capture-pane-contents` option. When you "save" with this option turned "on" the content of all the tmux panes is saved in files in `~/.tmux/resurrect/pane_contents/`. For some reason you don't have anything in that dir.

- What do you have in `~/.tmux/resurrect/pane_contents/` directory? Any files there?
- Did you try just restoring with `@resurrect-capture-pane-contents` set to "on" without previously saving the whole environment?
- can you try setting `set -g @resurrect-capture-pane-contents 'on'` (make sure this option is sourced), then \*save* and \*restore*. What happens? Still getting that error?

Limeth:
`~/.tmux/resurrect/pane_contents/` is empty, even after I save the session by pressing `<prefix><C-s>`.
I already have `set -g @resurrect-capture-pane-contents 'on'` in `~/.tmux.conf`.

My tmux version is 2.3, built from source.

sariattila:
Hi. Same issue here
![selection_002](https://cloud.githubusercontent.com/assets/5034149/15038645/c4e93b80-12a5-11e6-9060-46a02bf072e4.png)

passcod:
Same issue.

passcod:
The `pane_contents.tar.gz` file seems to contain the right things though.

ylluminarious:
+1, Also seeing the same as what @passcod said:

> The `pane_contents.tar.gz` file seems to contain the right things though.

[...]

PR #193, opened 2017-05-07
remove pane contents cleanup to repair restore

Mellbourn:
as suggested in https://github.com/tmux-plugins/tmux-resurrect/issues/141


bruno-:
Thanks

PR #194, opened 2017-05-07
remove terminal clearing since it collides with capture-pane-contents

Mellbourn:
The clearing of the terminal after history restore is not a feature. 
It conflicts with `resurrect-capture-pane-contents`. 
If you have both history restoration and content capture turned on, the screen will be restored then promptly erased.


bruno-:
I think this was added so that the `history` command is not seen in the command line. I agree the 'pane contents' feature is more important.

PR #197, opened 2017-06-01
add hostname to the path for saving and restoring sessions

comaniac:
This change is used for supporting saving and restoring sessions for the NFS on different servers.


bruno-:
Hey, thanks for the contribution.

I enabled this same feature in a different way. Please refer to the updated doc for changing the resurrect dir name https://github.com/tmux-plugins/tmux-resurrect/blob/master/docs/save_dir.md

PR #202, opened 2017-07-11
Changed the resurrect file path to allow it to work with Microsoft WLS

RogerioDosSantos:
The Microsoft Windows Linux Subsystem does not accept ":" on the file path. I did a small modification so the resurrect could work with Microsoft WLS. The side effect of this change is that is a user already have a resurrect file saved, it would have to save it again.


dc-mak:
I've run into the same problem - would appreciate/be glad to help in speeding this along.

JBlond:
Due https://github.com/tmux-plugins/tmux-resurrect/pull/272 this is obsolete

PR #206, opened 2017-09-24
Add instruction for NodeJS related project and tools

nguymin4:
Add instruction about restoring program related to NodeJS.

@bruno- There is a small issue which I am not so sure is a bug or not in the end of the documentation. Could you review it. Thank you.


bruno-:
Thank you.

PR #210, opened 2017-10-25
remove_old_backups() : replace xargs with -exec to handle 0 files found better

eugenevd:
If there are no files found by find due to mtime (ie, nothing older than 30 days), zero arguments are passed on to xargs' rm which result in 

Oct 25 14:14:50 xmachina save.sh[2772]: rm: missing operand
Oct 25 14:14:50 xmachina save.sh[2772]: Try 'rm --help' for more information.

to be seen in journalctl.

This way rm is only executed when something is found.


bruno-:
Thanks

PR #211, opened 2017-10-30
actually rename windows when restoring

malkomalko:


charlie-ht:
This was a show-stopping issue for my workflow, thanks very much for taking the time to create a PR!

thebearjew:
Big time bump on this. Desperately need windows to restore names!

jbleuzen:
Hi, would love to see this merged as well into tmux-resurrect... 
Tested it and it works perfectly !
Thanks

noscript:
Works for me as well!

ApricumMedia:
Yes, please, I find tmux window titles to be immensely helpful!

daizeng1984:
+1 wish to solve this issue. The original fix seems to break on newer tmux e.g. 2.6 due to asynchronic pane refreshing.

joelrobichaud:
Any reason this hasn't been merged yet? This is a showstopper for my workflow too.

peey:
This would be great to have. @bruno- @rummik could you review this and merge please? It just adds one line to the `restore.sh` file.

bruno-:
I think this solution will not be ideal for folks with `automatic-rename on`. It's possible we'll revert this addition because of that.

I think that was the original reason the PR was not merged early.

peey:
@bruno- ah I see your point.

Will the following be a correct solution? :

1. Keep track of status of the variable `automatic-rename` in [scripts/save.sh](https://github.com/tmux-plugins/tmux-resurrect/blob/master/scripts/save.sh#L36). If it's "off" during save then that means the name was set manually, if it's "on" then it means automatic-renaming was in effect.
2. During restore, add an if condition around the line introduced in this PR which checks if `automatic-rename` was on or off.

bruno-:
Maybe we don't need to track `automatic-rename` during save. Maybe just doing this **during restore** would work:

```
if automatic_rename_off
  rename_window
```

I'm not sure if that covers all the cases, but it's an improvement.

peey:
Current behaviour of window-level variable 'automatic-rename' seems to be that if it's set to "on", then when you set the window name, it's automatically turned "off" for that window. [source](https://superuser.com/q/791638/366614)

I think that's why you'll need to track it during the save.

peey:
@malkomalko would you be able to update this PR to take care of the `automatic-rename` issue?

It'll be definitely sad if this were to be reverted.

PR #216, opened 2017-11-19
Fix printing of command for ArchLinux

cedricpim:
On 4.13.12-1-ARCH, `uname -s` returns "Linux" and to get the correct
output of the command related to a PID, it requires the same options on
echo as OpenBSD and FreeBSD.


cedricpim:
Great, thanks :+1:

PR #230, opened 2018-01-19
Save and restore hooks

noscript:
Adds two for save restore events, that could be used to e.g. save and restore terminal window geometry, like so:

```
set -g @resurrect-save-hook 'eval $(xdotool getwindowgeometry --shell $WINDOWID); echo 0,$X,$Y,$WIDTH,$HEIGHT > $HOME/.tmux/resurrect/geometry'
set -g @resurrect-restore-hook 'wmctrl -i -r $WINDOWID -e $(cat $HOME/.tmux/resurrect/geometry)'
```

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -344,6 +344,9 @@ restore_active_and_alternate_sessions() {
 main() {
 	if supported_tmux_version_ok && check_saved_session_exists; then
 		start_spinner "Restoring..." "Tmux restore complete!"
+		if [ -n "$(restore_hook)" ]; then
+			eval "$(restore_hook)"
+		fi
 		restore_all_panes
 		restore_pane_layout_for_each_window >/dev/null 2>&1
 		if save_shell_history_option_on; then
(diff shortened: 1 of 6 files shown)


scripts/save.sh
```diff
@@ -325,6 +325,9 @@ main() {
 			stop_spinner
 			display_message "Tmux environment saved!"
 		fi
+		if [ -n "$(save_hook)" ]; then
+			eval "$(save_hook)"
+		fi
```
ashb: This should be just after the `save_all`

scripts/restore.sh
```diff
@@ -344,6 +344,9 @@ restore_active_and_alternate_sessions() {
 main() {
 	if supported_tmux_version_ok && check_saved_session_exists; then
 		start_spinner "Restoring..." "Tmux restore complete!"
+		if [ -n "$(restore_hook)" ]; then
+			eval "$(restore_hook)"
+		fi
```
ashb: Tricksy one - I would like a hook that is run for each pane. Once you add one hook people quickly want a hook everywhere, eh?

PR #233, opened 2018-01-24
Fix https://github.com/tmux-plugins/tmux-resurrect/issues/176.

### LINKED ISSUE #176: save with zoomed window leads to flickering
Wenn I zoomed into a pane and trigger save it zooms out and back in again. This is annoying if you use tmux-continuum, too, because the pane flickers every couple of minutes.

woutgg:
This removes (un)zooming on save which should not be necessary anymore with recent tmux versions, this should fix #176.
I think the relevant change in tmux itself has been made in [this commit](https://github.com/tmux/tmux/commit/531869bd92f0daff3cc3c3cc0ab273846f411dc8), so this change should work for tmux >=2.2.

Update: I have extended the tests a bit to also save and restore a zoomed pane.

### DIFF
--- a/scripts/save.sh
+++ b/scripts/save.sh
@@ -246,14 +246,8 @@ dump_windows() {
 			fi
 			# window_layout is not correct for zoomed windows
 			if [[ "$window_flags" == *Z* ]]; then
-				# unmaximize the window
-				toggle_window_zoom "${session_name}:${window_index}"
 				# get correct window layout
 				window_layout="$(tmux display-message -p -t "${session_name}:${window_index}" -F "#{window_layout}")"
-				# sleep required otherwise vim does not redraw correctly, issue #112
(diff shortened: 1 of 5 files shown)


scripts/save.sh
```diff
@@ -246,14 +246,8 @@ dump_windows() {
 			fi
 			# window_layout is not correct for zoomed windows
 			if [[ "$window_flags" == *Z* ]]; then
```
bruno-: Can't we just remove this whole `if` block? I can't see what value does a line that reassigns `window_layout` provides?

bruno-:
Tested and this works ok. Will merge after PR feedback is addressed (or clarified). Thanks for the PR

woutgg:
You are right, I was a bit too quick with the initial fix.

Issue #253, opened 2018-05-28
Programs are not being restored if server is killed


julenpardo:
The following works fine for me:

* Start the tmux server.
* Start Vim and other processes specified with `@resurrect-processes` option.
* Save session `prefix - C-s`
* Start a new tmux session
* Restore session

If I kill the tmux server after saving the session and try to make a restore in a new one, the panes are restored properly, but without the programs.

I'm using tmux 2.7 and zsh 5.1.1.

PR #256, opened 2018-05-29
#253: Get children process running within bash 

julenpardo:
Okay, after a little research, I understood what's going on and was able to make a workaround for this.

The restore was working fine because in `scripts/process_restore_helpers.sh`, in the `_process_should_be_restored`, we were in the first scenario, as the comment says, `where pane existed before restoration, so we're not restoring the process either`. So this is fine since the processes are not actually restored.

The problem comes when the processes have to be restored, caused by the txt restore files saved in `~/.tmux/restore`. The saved command is always `:-zsh`, saving the running children in a new line:

```
pane	1	1	:zsh	1	:*	0	:/home/jpardo/.tmux/plugins/tmux-resurrect	0	vim	:-zsh
vim -S
pane	1	1	:zsh	1	:*	1	:/home/jpardo/git/tmux-resurrect	1	zsh	:-zsh
pane	1	2	:ssh	0	:	0	:/home/jpardo	0	ssh	:-zsh
ssh h2601438.stratoserver.net
pane	1	2	:ssh	0	:	1	:/home/jpardo/.tmux/plugins/tmux-resurrect	1	zsh	:-zsh
pane	1	3	:zsh	0	:-	0	:/home/jpardo	0	mysql	:-zsh
mysql -uroot -p
pane	1	3	:zsh	0	:-	1	:/home/jpardo/.tmux/resurrect	1	zsh	:-zsh
pane	1	4	:python3	0	:	0	:/home/jpardo	1	python3	:-zsh
python3
pane	1	5	:php	0	:	0	:/home/jpardo	1	php	:-zsh
php -a
pane	main	1	:zsh	1	:*	0	:/home/jpardo	1	zsh	:-zsh
window	1	1	1	:*	850a,211x61,0,0[211x34,0,0,2,211x26,0,35,3]
window	1	2	0	:	9942,211x61,0,0{105x61,0,0,4,105x61,106,0,5}
window	1	3	0	:-	0d95,211x61,0,0[211x28,0,0,6,211x32,0,29,7]
window	1	4	0	:	b365,211x61,0,0,8
window	1	5	0	:	b366,211x61,0,0,9
window	main	1	1	:*	59df,211x61,0,0,10
state	1	main
```

The `vim`, `ssh`, etc. commands are saved in a new line.

I modified the `save_command_strategies/ps.sh` to tail the last line of the `ps` output, so the resulting `last` with this workaround would be:

```
pane	1	1	:zsh	1	:*	0	:/home/jpardo/.tmux/plugins/tmux-resurrect	0	vim	:vim -S
pane	1	1	:zsh	1	:*	1	:/home/jpardo/git/tmux-resurrect	1	zsh	:-zsh
pane	1	2	:ssh	0	:-	0	:/home/jpardo	0	ssh	:ssh h2601438.stratoserver.net
pane	1	2	:ssh	0	:-	1	:/home/jpardo/.tmux/plugins/tmux-resurrect	1	zsh	:-zsh
pane	1	3	:zsh	0	:	0	:/home/jpardo	0	mysql	:mysql -uroot -p
pane	1	3	:zsh	0	:	1	:/home/jpardo/.tmux/resurrect	1	zsh	:-zsh
pane	1	4	:python3	0	:	0	:/home/jpardo	1	python3	:python3
pane	1	5	:php	0	:	0	:/home/jpardo	1	php	:php -a
pane	main	1	:zsh	1	:*	0	:/home/jpardo	1	zsh	:-zsh
window	1	1	1	:*	850a,211x61,0,0[211x34,0,0,2,211x26,0,35,3]
window	1	2	0	:-	9942,211x61,0,0{105x61,0,0,4,105x61,106,0,5}
window	1	3	0	:	0d95,211x61,0,0[211x28,0,0,6,211x32,0,29,7]
window	1	4	0	:	b365,211x61,0,0,8
window	1	5	0	:	b366,211x61,0,0,9
window	main	1	1	:*	59df,211x61,0,0,10
state	1	main
```

Which allows the successful restore of the processes.

### DIFF
--- a/save_command_strategies/ps.sh
+++ b/save_command_strategies/ps.sh
@@ -22,7 +22,8 @@ full_command() {
 	ps "$(ps_command_flags)" "ppid command" |
 		sed "s/^ *//" |
 		grep "^${PANE_PID}" |
-		cut -d' ' -f2-
+		cut -d' ' -f2- |
+		tail -1
 }
 
 main() {


bruno-:
Do you know the root reason why the output comes with a newline? It's the first time I see this kind of issue.

julenpardo:
I think that the reason is that two processes are being created in the pane. This is how I concluded it:

- I get the active PIDs:
```
tmux list-panes -a -F '#{pane_pid} #{pane_pid}'
```

Which returns:
```
3777 %1
8629 %2
```

- The `3777` process is a zsh instance:

```
ps -o cmd fp 3777
```

Returning 

```
CMD
-zsh
```

- I look for its children processes:

```
pgrep -P 3777
```

Having two of them:

```
4055
8510
```

Being the first one another zsh process, and the second one the process we actually want to save (vim, ssh, etc.)

So the saving strategy gets the commands of those two PIDs.

I would bet that that's the reason of why the command comes in a newline. But I don't know the reason of why there are being created two shell processes.

bruno-:
Thank you for the analysis. As you hinted, it's unusual that `pgrep -P <pid>` returns 2 processes.. one of them being another zsh process.

Do you maybe have a zsh plugin that would create that?

julenpardo:
Could be. I'll run zsh without any plugin or config file and let you know about it.

PR #272, opened 2018-09-10
Windows file names can not contain a colon

JBlond:
That would solve https://github.com/tmux-plugins/tmux-resurrect/issues/131

### DIFF
--- a/scripts/helpers.sh
+++ b/scripts/helpers.sh
@@ -112,7 +112,7 @@ _RESURRECT_DIR="$(resurrect_dir)"
 
 resurrect_file_path() {
 	if [ -z "$_RESURRECT_FILE_PATH" ]; then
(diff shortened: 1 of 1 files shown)


bruno-:
I'm in favor of merging this, but I'm not sure if we want to use this format `%Y-%m-%dT%H-%M-%S`. The previous one was following ISO8601... I'm not sure if the new hyphen-only format follows any standard?

Based on the [ISO8601 Times section](https://en.wikipedia.org/wiki/ISO_8601#Times), it seems colon or nothing are the options?

After skimming the wikipedia ISO8601 page, I think we could go with this format `%Y%m%dT%H%M%S`?

JBlond:
@bruno- Are you going to merge this after I changed the PR?

bruno-:
@JBlond thanks for the patience. Due to the amount of email/work I just forget about things. Thanks for reminding me.

PR #279, opened 2018-11-15
fix: pane_content not being restored

BarbzYHOOL:
Taken from this Mellbourn@0a6f907 and this tmux-plugins#141

Closes tmux-plugins#141

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -258,9 +258,6 @@ restore_all_panes() {
 			restore_pane "$line"
 		fi
 	done < $(last_resurrect_file)
-	if is_restoring_pane_contents; then
-		rm "$(pane_contents_dir "restore")"/*
-	fi
 }
 
 restore_pane_layout_for_each_window() {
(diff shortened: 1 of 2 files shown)


bruno-:
Thanks for the fix. I see it handles #141, but would you mind explaining why it works?

BarbzYHOOL:
I had only a quick look, I wanted to fix it myself but I saw that someone else did it so not much analyse has been put into this from my part lol.

Basically, I think that this part is enough

```
if is_restoring_pane_contents; then
		pane_content_files_restore_from_archive
fi
```

And the part I removed was just "overriding" the code above

```
if is_restoring_pane_contents; then
		rm "$(pane_contents_dir "restore")"/*
fi
```

I looked at the pane_contents/ dir and it's always empty so I supposed it was extracted from the .gz during restore, then removed at the end of the restore (so it always looks empty). Maybe this part of the code removed the content too early.


Sorry, can't say much because I have no time to analyze exactly why it works, it's mostly suppositions.

Note that the guy who found the "fix" first said it was safe to use

PR #281, opened 2018-12-10
Fix #94 start default-shell as login shell.

### LINKED ISSUE #94: Sourced bash_profile is lost upon restore
I may be doing something wrong here, and please share some wisdom if that is the case. I keep quite a number of handy alias's in my bash_profile, along with a few env variables. When I start up tmux and build my environment, I have no problems, but when I resurrect I lose my sourced bash_profile, as well as a few tmux.conf items such as colors etc. Thanks in advance for any help!

lwilke:
If pane content is restored and no default-command is set, restore.sh
falls back to default-shell, which tmux sets automatically and calls as
a login shell.

In pane_creation_command() we need to make sure to call default-shell as a login shell.
As tmux does. But take care not to call default-command as a login shell.

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -100,7 +100,7 @@ tmux_socket() {
 # tmux server each time.
 cache_tmux_default_command() {
 	local default_shell="$(get_tmux_option "default-shell" "")"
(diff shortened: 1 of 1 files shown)


bruno-:
Thanks

PR #283, opened 2018-12-28
Add cmdline save strategy

### LINKED ISSUE #60: [Bug] watch command resurrects without '-characters.
First of all, thank you for making this really awesome tool that I've been waiting for a long time now!
Also, I'm pretty new on Git(Hub) and as such I'm not sure how to report new issues (or feature requests for that matter). If you have a comment then please let me know how I could do things better next time :)

On to the situation:

After I've saved the session (only 1) and I tried to resurrect it, I found that in one of the panes the command `watch -n1 -d 'free -m; df -h'` was turned into `watch -n1 -d free -m; df -h` upon resurrection.

Also I've found that the htop pane had more space than previously, but I suppose that's material for another bug report?

Here's the [~/.tmux/resurrect/last and my .tmux.conf](https://gist.github.com/absorber/b4cbfcb560ea221633c0)

rummik:
This *should* offer a fix for #60 on the Linux-side of things.  Though, it is platform specific.

The `-r` parameter to `read` just tells it not to interpret backslash escapes, which is probably what should be used anyway.


scripts/restore.sh
```diff
@@ -293,9 +293,10 @@ restore_shell_history() {
 
 restore_all_pane_processes() {
 	if restore_pane_processes_enabled; then
+		local save_command_strategy="$(get_tmux_option "$save_command_strategy_option" "$default_save_command_strategy")"
```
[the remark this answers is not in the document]
rummik: Oh.  I forgot to remove that line after removing the dependency on `base64`.  There was an issue where it would potentially have broken other restore strategies if the command could be read as a base64 encoded string, which is why that line was included

save_command_strategies/cmdline.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+CURRENT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+PANE_PID="$1"
+CPID=$(pgrep -P $PANE_PID)
```
bruno-: What does `CPID` stand for? Can we use expanded name?

save_command_strategies/cmdline.sh
```diff
@@ -0,0 +1,23 @@
...
+
+exit_safely_if_empty_ppid() {
+	if [ -z "$PANE_PID" ]; then
+		exit 0
+	fi
+}
+
+full_command() {
+	[[ -z "$CPID" ]] && exit 0
+	cat /proc/${CPID}/cmdline | xargs -0 printf "%q "
```
bruno-: Since this strategy uses `procfs` I think a more fitting name for it would be `procfs` or maybe even `linux_procfs` to indicate it's linux only. I suggest renaming this file to `save_command_strategies/linux_procfs.sh`

bruno-: I would expect `/proc/${CPID}/cmdline` to output a single line.. so can you please clarify why we need `| xargs -0 printf "%q "`?

rummik: It contains a null-terminated set of strings representing each argument, which are then separated into individual arguments by `xargs` and passed to `printf` to be quoted.  This makes it so arguments with spaces are preserved

rummik: I think that's pretty reasonable; I'll go ahead and do that later today

save_command_strategies/cmdline.sh
rummik: Child PID

PR #284, opened 2018-12-29
Add mosh-client default strategy

rummik:
This also includes changes to `restoring_programs.md` in order to let people
know to use `mosh-client` rather than `mosh`


bruno-:
Thanks

PR #285, opened 2018-12-29
Some fixes for zsh history save/restore

rummik:
I noticed some of the keybindings for zsh either weren't set or had values that
gave TMUX some trouble.  This PR should fix that.


bruno-:
Hey,
I'm not in favor of merging this feature. It's hacky and improves an experimental feature that is already pretty dirty.

If you want, you can add `zsh` instructions to this file `docs/restoring_shell_history.md`

PR #286, opened 2018-12-29
Add @resurrect-process-match-strategy option and "basename" strategy

rummik:
This adds a process matching strategy that functions similarly to running
`basename` on the program path

It can be enabled with

	set -g @resurrect-process-match-strategy 'basename'


bruno-:
I think this can already be achieved via `~`, see here: https://github.com/tmux-plugins/tmux-resurrect/blob/master/docs/restoring_programs.md#general-instructions-

If have a process, `/usr/bin/bla/process` it can be matched with just `~process`.

rummik:
The `~` prefix matches any part of the command string, which can result in restoring commands that happen to have an argument that contains the text being matched

PR #289, opened 2019-02-08
Updating resurrect.tmux to be capable to use links as run-shell sources.

HenryGiraldo:
Right now you can't use symbolic links the bottom of .tmux.conf.

> run-shell ~/clone/path/resurrect.tmux

Adding changes to allow it.


HenryGiraldo:
Hi, have you considered to add this change?

I have the configuration files in a repository and I link them at users home. This type of setup needs this change to work. I think it is quite powerful and the change is minimal.

PR #295, opened 2019-03-27
Proposal to hide the input commands saving/restoring bash history:

Gullumluvl:
I configured my tmux to use resurrect save on each detach, but with the option `resurrect-save-shell-history` ON, this creates a lot of useless "history" lines in my panes. So I created a configuration to hide those commands. It's super experimental, uses unchecked key mappings, so I make this pull request for the reference, but this might need improvement, or another implementation idea.

First, the following must be configured in your ~/.bashrc:
bind -x '"\C-[z": tput smcut'  # Enter a secondary terminal buffer with Alt-z
bind -x '"\C-[Z": tput rmcut'  # Quit it with Alt-Z

Then the tmux sent keys can include those controls around the executed
commands to invisibly save/restore the history.

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -281,7 +281,7 @@ restore_shell_history() {
 
 				if [ "$pane_command" = "bash" ]; then
 					local read_command="history -r '$history_file'"
-					tmux send-keys -t "$pane_id" "$read_command" C-m
+					tmux send-keys -t "$pane_id" M-z "$read_command" C-m M-Z
 				elif [ "$pane_command" = "zsh" ]; then
 					local accept_line="$(expr "$(zsh -i -c bindkey | grep -m1 '\saccept-line$')" : '^"\(.*\)".*')"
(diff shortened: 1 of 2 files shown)


fruch:
I'm on ubuntu with the default terminal and this isn't working:
I'm getting the following error
```
tput: unknown terminfo capability 'smcut'
```

fruch:
@Gullumluvl I like it, I wish it could be working with definitions only in `.tmux.conf` and not `.bashrc`

eventfully I've disable the style changes on bell in my setup, it was too much having all my windows flash every time this command runs

Gullumluvl:
My bad, I should have just copy-pasted... I didn't notice the flickering indeed. Thanks for your feedback, I would also like to have all config in one place but could not figure it out...

fruch:
Also one this that's annoying with this, is if you type while this kicks in, it messes you current command you are typing, two times this week it happened to me and had to blindly type `reset` to fix.

Takes a few seconds to figure what's going on

Issue #179, opened 2017-01-18
tmux-resurrect commands through tmux in -CC session


jhgorse:
Greetings!

I run tmux -CC in iTerm2 and I do not have access to the usual key bindings via prefix (C-b) keys in the -CC session windows.

How else can I interact with tmux-resurrect to save/restore the session?

bruno-:
Hi,

while I experimented with tmux + iTerm2 I must admit I never understood what it is about.. so this advice might or might not work:

tmux-resurrect key bindings just call plain old bash scripts. You should be able to manually run these scripts with:

- for save: `~/.tmux/plugins/tmux-resurrect/scripts/save.sh`
- for restore: `~/.tmux/plugins/tmux-resurrect/scripts/restore.sh`

gnachman:
From the issue report on iTerm2 (https://gitlab.com/gnachman/iterm2/issues/7049) it looks like tmux-resurrect must be running tmux with `-CC`.

NishantTharani:
This seems to work for me in a hacky way. What I do:

1. Run `tmux -CC new -s myname` and do stuff
2. Run `~/.tmux/plugins/tmux-resurrect/scripts/save.sh`
3. Detach from the tmux session and then kill it
4. Run `tmux -CC new -s myname` again
5. Run `~/.tmux/plugins/tmux-resurrect/scripts/restore.sh`
6. An error window pops up in iTerm, press 'Okay'
7. The tmux window that was created in step 4 will have now vanished. Run `tmux -CC attach` and it'll come back, restored correctly

TimothyBramlett:
@NishantTharani That doesn't seem to work for me, strangely.

PR #297, opened 2019-04-03
Corrected what is the expected value of the $CURRENT_DIR

delphym:
It is necessary for the cases when running the scripts (as symlinks) without a
possibility of using key bindings as advised in issue #179
https://github.com/tmux-plugins/tmux-resurrect/issues/179

The case is e.g. saving the sessions for byobu and having the symlinks
created at $HOME/bin/ as:
byobu-session-restore ->
/home/delphym/.tmux/plugins/tmux-resurrect/scripts/restore.sh

byobu-session-save ->
/home/delphym/.tmux/plugins/tmux-resurrect/scripts/save.sh

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -1,7 +1,16 @@
 #!/usr/bin/env bash
 
-CURRENT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+# Absolute path to this script, e.g. /home/user/.tmux/plugins/tmux-resurrect/scripts/save.sh
+SCRIPT=$(readlink -f "$0")
+# Absolute path this script is in, thus /home/user/.tmux/plugins/tmux-resurrect/scripts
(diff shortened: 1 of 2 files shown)


delphym:
This is what it looks like when having a symlink in ~/bin
ζ byobu-session-restore
/home/delphym/bin/byobu-session-restore: line 5: /home/delphym/bin/variables.sh: No such file or directory
/home/delphym/bin/byobu-session-restore: line 6: /home/delphym/bin/helpers.sh: No such file or directory
/home/delphym/bin/byobu-session-restore: line 7: /home/delphym/bin/process_restore_helpers.sh: No such file or directory/home/delphym/bin/byobu-session-restore: line 8: /home/delphym/bin/spinner_helpers.sh: No such file or directory
/home/delphym/bin/byobu-session-restore: line 347: supported_tmux_version_ok: command not found

Ideal seems to rename the $CURRENT_DIR variable, which I did at the first instance, but then it was still complaining on to execute l. 51 from check_tmux_version.sh

PR #300, opened 2019-06-17
Add support for linked windows

pokey:
Correctly save / load sessions which have used the `tmux link-window` command


mbaltrusitis:
Applied this patch and works as expected.

PR #307, opened 2019-08-27
Add resurrect-limit-shell-history option

marc-orr:
This patch adds a new tmux-ressurect option,
resurrect-limit-shell-history, to limit the frequency of saving shell
history to once per day. The motivation of this option is that otherwise
resurrect is run much more frequently (e.g., ~ every 15 minutes), and
saving the history spams the console (saving other state does not...).
Spamming the console ~once per day, while still saving history, seems
like a reasonable tradeoff---hence this patch.


bruno-:
Hey, I'm sorry you invested time and effort in making this work, but "restore shell history" is long-time unsupported. I unofficially mentioned it's unsupported and deprecated in issues. My bad for not deprecating it officially - I intend to that that this week https://github.com/tmux-plugins/tmux-resurrect/pull/308

Issue #288, opened 2019-02-03
tmux shell spammed with... history -r '/.tmux/resurrect/bash_history-0:0.1' 


KayHS:
how do i stop my tmux shell from being spammed with..

history -w '/Users//.tmux/resurrect/bash_history-0:0.1'
history -r '/.tmux/resurrect/bash_history-0:0.1'

k3dar:
i think you have enable "Restoring shell history" and this is command for save and restory history for each tmux (sub)windows...

you can disable it removing line "set -g @resurrect-save-shell-history 'on'" from your .tmux.conf

more info: https://github.com/tmux-plugins/tmux-resurrect/blob/master/docs/restoring_shell_history.md

Bubblemelon:
Having the same problem with being spammed:
> history -w '/home/cherylfong/.tmux/resurrect/bash_history-0:0.0'
history -r '/home/cherylfong/.tmux/resurrect/bash_history-0:0.0'

Tried @k3dar 's solution by adding `PROMPT_COMMAND="history -a; $PROMPT_COMMAND"` to `.bashrc` but this did not resolve the issue.

k3dar:
> Having the same problem with being spammed:
> 
> > history -w '/home/cherylfong/.tmux/resurrect/bash_history-0:0.0'
> > history -r '/home/cherylfong/.tmux/resurrect/bash_history-0:0.0'
> 
> Tried @k3dar 's solution by adding `PROMPT_COMMAND="history -a; $PROMPT_COMMAND"` to `.bashrc` but this did not resolve the issue.

this is only different way to store command history, but for despamming you also need:
"disable it removing line "set -g @resurrect-save-shell-history 'on'" from your .tmux.conf"

k3dar:
> I just mocked up a local patch to save the history once per day

but with this patch/sollution you lost last <=23:59 history if you os crash/reboot/poweroff... 
with change i write [above](https://github.com/tmux-plugins/tmux-resurrect/issues/288#issuecomment-473608158) is still history save immediately after enter (respectively after entered command finish/exit)

virusdave:
>  all (sub)windows have complete history from all previous (sub)windows...

This is very undesirable, and is why i (and i suspect most others) use this in the first place.  We/I want each window's pane's histories tracked separately.  At least i think i do :)  Ask me again after the first dozen times i accidentally close the wrong pane.  Then i might want a shared global history.  But for now, i want them separate.

[...]

PR #308, opened 2019-08-27
Deprecate restoring shell history

bruno-:
"restoring shell history" feature is very invasive and dirty. I propose
removing it in the future without the replacement.

I worry it spoils the first experience of using this plugin for users
that accidentally turn it on and don't know how to turn it off, see
https://github.com/tmux-plugins/tmux-resurrect/issues/288

Also, it hurts me to reject PRs that improve the feature.

Thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - delete resurrect files older than 30 days, but keep at least 5 files
 - add save and restore hooks
 - always use `-ao` flags for `ps` command to detect commands
+- Deprecate restoring shell history feature.
 
 ### v2.4.0, 2015-02-23
 - add "tmux-test"
--- a/README.md
+++ b/README.md
@@ -49,7 +49,6 @@ Optional:
 
 - [restoring vim and neovim sessions](docs/restoring_vim_and_neovim_sessions.md)
 - [restoring pane contents](docs/restoring_pane_contents.md)
-- [restoring shell history](docs/restoring_shell_history.md) (experimental)
 
 Requirements / dependencies: `tmux 1.9` or higher, `bash`.
 
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -20,7 +20,7 @@ Currently the following hooks are supported:
 
   Called before any tmux state is altered.
 
-- `@resurrect-hook-pre-restore-history`
+- `@resurrect-hook-pre-restore-history` - deprecated
 
   Called after panes and layout have been restores, but before bash history is
   restored (if it is enabled) -- the hook is always called even if history
--- a/docs/restoring_shell_history.md
+++ b/docs/restoring_shell_history.md
@@ -1,4 +1,8 @@
-# Restoring shell history (experimental)
+# Restoring shell history (deprecated, do not use)
+
+This feature is deprecated because it's very invasive. It will be removed in
+the future with no replacement. To see problems it causes check
+[this issue](https://github.com/tmux-plugins/tmux-resurrect/issues/288).
 
 **Supported shells**: `bash` and `zsh`.
 
--- a/scripts/variables.sh
+++ b/scripts/variables.sh
@@ -38,7 +38,7 @@ pane_contents_area_option="@resurrect-pane-contents-area"
 default_pane_contents_area="full"
 
 bash_history_option="@resurrect-save-bash-history" # deprecated
-shell_history_option="@resurrect-save-shell-history"
+shell_history_option="@resurrect-save-shell-history" # deprecated
 
 # set to 'on' to ensure panes are never ever overwritten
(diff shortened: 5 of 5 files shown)


bruno-:
I'm not sure. Maybe try something with the hooks feature:

https://github.com/tmux-plugins/tmux-resurrect/blob/ca6468e2deef11efadfe3a62832ae67742505432/scripts/restore.sh#L373

I'd be open to adding another hook if it helped you.

rbren:
Adding this to my `.bashrc` seems to do the trick. I'm sure there's an edge case I'm not thinking of because it's dead simple. Adapted from another issue I found in this repo

```bash
HISTS_DIR=$HOME/.bash_history.d
mkdir -p "${HISTS_DIR}"

if [ -n "${TMUX_PANE}" ]; then
  PANE=$(echo $TMUX_PANE | sed 's/%//g')
  export HISTFILE="${HISTS_DIR}/bash_history_tmux_${PANE}"
else
  export HISTFILE="${HISTS_DIR}/bash_history_no_tmux"
fi
```

bruno-:
I'm not sure if `$TMUX_PANE` numbers are consistent across `tmux-resurrect` restores.

But I'm sure you can work around that by using something like this:

```
${session_name}:${window_number}:${pane_index}
```

Of course, you'll have to get each id "component" separately via some `tmux` subcommand.

If you get it working consider creating a short instruction in `docs/`.

rbren:
$TMUX_PANE has been pretty consistent, but I do think I saw it fail at least once. Here's a second attempt:

```bash
pane_id_prefix="resurrect_"

# Create history directory if it doesn't exist
HISTS_DIR=$HOME/.bash_history.d
mkdir -p "${HISTS_DIR}"

if [ -n "${TMUX_PANE}" ]; then
  pane_id=$(tmux display-message -p '#{pane_title}')
  if [[ $pane_id != "$pane_id_prefix"* ]]; then
    random_id=$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 16)
    printf "\033]2;$pane_id_prefix$random_id\033\\"
    pane_id=$(tmux display-message -p '#{pane_title}')
  fi
  export HISTFILE="${HISTS_DIR}/bash_history_tmux_${pane_id}"
else
  export HISTFILE="${HISTS_DIR}/bash_history_no_tmux"
fi
```

To sum up - if the `pane_title` hasn't already been set, set it to `resurrect_$RANDOM`, and use that as the identifier for `$HISTFILE`. `pane_title` seems to be preserved well in save/restore.

The downside is that you can't customize pane titles, unless you use the same prefix as above (and even then renaming will wipe out your history). There's also the narrow chance of a collision.

If you think this is a decent solution I'm happy to submit a PR to `/docs`

bruno-:
Can't two panes (in different windows or in different sessions) have the same title?
And yea, if a person is actually using pane titles and actively changing them this won't work.

What would be the downside of using `tmux display-message -p "#{session_name}:#{window_index}:#{pane_index}"` for the identifier?

rbren:
The issue with `pane_index` is that it changes as you move windows around. E.g. if you have a vertical split, and then swap the left and right panes, you also swap their `pane_index`, so you'd also swap history.

I thought about trying to listen for changes to `pane_index` but that seems pretty error-prone.

And yes - this solution would force each pane to have a unique title by using a random identifier.

bruno-:
If you want to do it I'm ok with submitting a PR to `docs/`.

I propose clearly mentioning the downsides of this approach and maybe even ask for help? I wouldn't be surprised if someone came up with a better approach.

jamie01:
Thanks for this trick, seems like a fairly neat solution. I for one never use pane titles so the loss of them isn't a big deal.

I chose to modify the name generator to use the current epoch timestamp plus 1 random character, for cases where multiple panes get created within a second. This has the nice benefit to me of being clear which panes are older than others. Not sure I need that but it seemed convenient to throw it in while at it.

So my `random_id=$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 16)` becomes
`printf -v random_id %s $(date +%s) $(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 1)`

seamusdemora:
Would it help to keep a separate history for each tmux session, and then add it to  the overall (e.g.) `bash` history when the session was deleted?

barbosa46:
When you create a session and it loads your bashrc/zshrc, for some reason, the pane_id is set to your hostname. I made this shortcut to source it again when creating a new session. If you know of a more ellegant way to deal with this, please tell me.
```bash
function tms(){
    tmux new-session -d -s $1 'source $HOME/.zshrc';
    tmux split-window;                             # split the detached tmux session
    tmux send 'source $HOME/.zshrc' ENTER;                     
    tmux send 'clear' ENTER;                     
    tmux a;
}

```

Issue #192, opened 2017-05-06
cat: /Users/.tmux/resurrect/pane_contents//pane-jcs:5.1: No such file or directory


eternalfool:
I get

`cat: /Users/shashwat/.tmux/resurrect/pane_contents//pane-jcs:5.1: No such file or directory`

in every window of the when I resurrect. Also none of the contents of the pane are resurrected.

bruno-:
Hm.. this feature should work nicely. It's been out for years now and worked for me personally without problems.

Would you mind providing your `tmux.conf`?

Did you try the usual stuff: completely restarting tmux after installing the plugin and/or changing some tmux-ressurect options?

kentmartin73:
Hi,

Weird, I get this too:

> cat: '/home/xxxx/.tmux/resurrect/restore/pane_contents//pane-0:0.3': No such file or directory

tmux.conf is as follows:

```
run-shell "source /usr/local/lib/python3.5/dist-packages/powerline/bindings/zsh/powerline.zsh"
# run-shell "powerline-daemon -q"

# source /usr/local/lib/python3.5/dist-packages/powerline/bindings/tmux/powerline_tmux_2.1_plus.conf
source /usr/local/lib/python3.5/dist-packages/powerline/bindings/tmux/powerline.conf

set -g mouse on
# to enable mouse scroll, see https://github.com/tmux/tmux/issues/145#issuecomment-150736967
# run-shell ~/.tmux-better-mouse-mode/scroll_copy_mode.tmux
set -ga terminal-overrides ',:xterm*:smcup@:rmcup@'

# List of plugins
set -g @plugin 'tmux-plugins/tpm'
# Needed to make copy clipboard work on OSX (on some versions)
# set-option -g default-command "reattach-to-user-namespace -l zsh"

set -g @plugin 'tmux-plugins/tmux-sensible'
set -g @plugin 'tmux-plugins/tmux-yank'

# History searching controlled with / then n & N
set -g @plugin 'tmux-plugins/tmux-copycat'

# Save and resurrect sessions
set -g @plugin 'tmux-plugins/tmux-resurrect'
set -g @plugin 'tmux-plugins/tmux-continuum'
set -g @continuum-restore 'on'
set -g @continuum-save-interval '1'
set -g @resurrect-capture-pane-contents 'on'

# Clipboard handling
[...]

PR #313, opened 2019-10-18
Use 'cat' binary to restore pane contents rather than (potentially) using a shell command or alias

9999years:
`pane_creation_command` in `scripts/restore.sh` uses the string `cat` to run `/bin/cat`; however, because `pane_creation_command` is evaluated in the shell, aliases of `cat` may cause unexpected results; in my case, it runs the [`bat`][1] binary, which produces weird and garbled output. By using `command -v`, we can get the path of the `cat` binary and avoid this. (Note that we don't use `which` because `which` isn't POSIX-standard.)

[1]: https://github.com/sharkdp/bat

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -108,7 +108,7 @@ tmux_default_command() {
 }
 
 pane_creation_command() {
-	echo "cat '$(pane_contents_file "restore" "${1}:${2}.${3}")'; exec $(tmux_default_command)"
+	echo "$(command -v cat) '$(pane_contents_file "restore" "${1}:${2}.${3}")'; exec $(tmux_default_command)"
 }
 
 new_window() {


bruno-:
Hi,

I think overriding standard shell commands is an anti-pattern.

I'll leave this PR open in case the community disagrees, but I'm not in favor of merging this.

9999years:
Even if redefining coreutils commands is an anti-pattern, running a binary using its absolute name is the *correct* way to send commands to a user's shell; given that this patch has no cost — it makes `tmux-resurrect` work on more systems without doing anything unclear or strange in the code — I think we should merge it.

eproxus:
I have a gut feeling this will solve #192, or at least is related to it. When using Fish as a shell I always get this message in the last pane and no history:
```
cat: /Users/user/.tmux/resurrect/restore/pane_contents//pane-dev:3.1: No such file or directory
```

eproxus:
Just tested the patch and it seems it did not fix #192.

eproxus:
Just hit this problem today. Using Fish shell and setting an alias from `cat` to `bat` breaks the pane restoration. Is there any possibility to merge this? I assume lots of people could have `cat` aliases.

Issue #264, opened 2018-08-06
@restore-processes did not restore ssh process


rbhanot4739:
I was going through the documentation and saw this option to restore additional programs and ssh was one of them. So i tried but i was not able to restore the ssh sessions after i started the tmux after the reboot.  

All my panes with their contents were restored, but the pane having ssh didn't and failed with "no route to host" and this error is not really correct because i can easily login manually.
Below is how my setup looks in tmux.conf

set -g @plugin 'tmux-plugins/tmux-resurrect'                                      
set -g @resurrect-capture-pane-contents 'on'                                      
set -g @resurrect-processes 'ssh'                                                 
# Initialize TMUX plugin manager (keep this line at the very bottom of tmux.conf) 
run '~/.tmux/plugins/tpm/tpm'                                                     

I have keys setup properly so my ssh sessions don't really need the password.

k3dar:
for me work restoring ssh session ok, you can look to your ${resurrect-dir}/@last what is here save for ssh command, if your try run this saved manualy, also connect?

cocobear:
In my situation , I had ssh alias to another function , if I unset ssh alias, it worked.

spencerbutler:
Awesome plugin! Thanks!

I ran into this issue, and I found the cause, and a solution that works for me. I haven't read all of the code yet, but as far as I can tell, this change shouldn't negatively impact anyone else. I'll submit a PR for review.

### Problem
Restores to ssh panes only worked some of the time. As it turns out, the panes that didn't work[1] were invoked without a controlling terminal and their PIDs were not outputted with ps -ao "ppid command". 
### Solution
Change the full_command function to `ps -o cmd= "${PANE_PID}"`

[1] from man 1 tmux `bind-key S command-prompt "new-window -n %1 'ssh %1'"`

fbicknel:
As with @cocobear , I have `ssh` aliased to `ssh-ident`. For some reason, resurrect saves it as `/usr/bin/ssh`. I edited `~/.tmux/resurrect/last` changing `/usr/bin/ssh` to simply `ssh` (and removing the `-oUseRoaming=no` option as well), saved that, `<prefix>d` -> killed tmux -> start tmux -> `<prefix>^r` -> and it worked.

weilannw:
I alias my ssh to a script that runs autossh. By adding autossh instead of ssh to the list it correctly restores my sessions. It seems like adding the actual program to the list would address most of these cases as it did for me

[...]

PR #314, opened 2019-11-01
allow children of tmux to be saved

spencerbutler:
I noticed that your awesome plugin was not saving all my ssh sessions.  Although my use case was different than those in #264, I added a comment there with my issue. The following PR builds on what I originally posted there.

cheers

### DIFF
--- a/save_command_strategies/ps.sh
+++ b/save_command_strategies/ps.sh
@@ -4,6 +4,12 @@ CURRENT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 PANE_PID="$1"
 
+is_command_tmux_child() {
+  if [ $(ps -o cmd= $(ps -o ppid= "${PANE_PID}")) == tmux ]; then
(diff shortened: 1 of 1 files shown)


bruno-:
Based on what I understood from #264 are you saying you're running ssh directly from `tmux`?
Something like `$ tmux new-window 'ssh yourdomain.com'`?

If the above is true, then I'm not sure we want to support that. It's an edge case that adds complexity. The binding from #264 can easily be changed to: open new pane/window with a terminal + start ssh. With this change everything should work just fine.

spencerbutler:
I don't see this as an edge case. Not everyone spawns (or wants to spawn) a new shell for every new window. I also don't understand the complexity this introduces. You only support a handful of programs to restore, and this is added at the end of that discovery process before giving up.

Is there a less complex way I could include such child processes? Are there others that would appreciate restoring the children of tmux?

MattBystrin:
@bruno-,
I don't see this as an edge case too. It is very comfortable to open some app like this

PR #318, opened 2019-11-29
bash history restore

javiplx:
The changes basically replace the use of *pane_index* by *pane_id*. This guarantees that the environment variable `TMUX_PANE` is consistent across restores, and allows to use it as index to get separate history files for different panes.
I have this on my `.bashrc`, to set the right history filename in tmux and to update it on every command
```
HISTFILE=${HISTFILE}${TMUX_PANE:+.${TMUX_PANE#%}}
[ x$TMUX_PANE != x ] && PROMPT_COMMAND="${PROMPT_COMMAND:+$PROMPT_COMMAND$'\n'}history -a"
```
Main problem was restoring directory & content of the very first pane (`%0'), but I've included an alias for that.

### DIFF
--- a/resurrect.alias
+++ b/resurrect.alias
@@ -1,5 +1,5 @@
 
-alias resurrect='eval $( while read line_type session_name window_number window_name window_active window_flags pane_index dir pane_active pane_command pane_full_command; do
+alias restore='eval $( while read line_type session_name window_number window_name window_active window_flags pane_index dir pane_active pane_command pane_full_command; do
       [ x$pane_index = x0 ] && echo "cd ${dir#:} ; tmux new-session -s ${session_name#:} -n ${window_name} \" tar -xzf $HOME/.tmux/resurrect/pane_contents.tar.gz -C $HOME/.tmux/resurrect/restore --to-stdout ./pane_contents/pane-${session_name#:}:${window_number}.%${pane_index} ; exec ${pane_command:-/bin/bash}\" ; cd - > /dev/null"
     done < $HOME/.tmux/resurrect/last )'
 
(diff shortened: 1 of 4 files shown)


tessus:
@javiplx interesting. I'm using the following with git master and it works splendidly:

```
[[ "$TMUX" ]] && TSWP=`tmux display -pt "${TMUX_PANE:?}" '#{session_name}:#{window_index}.#{pane_index}'` && HISTFILE=$HOME/.bash_history.d/bash_history-$TSWP
PROMPT_COMMAND="history -a; $PROMPT_COMMAND"
```

My histories are always restored properly. Just wondering why this change would make this any better.

javiplx:
> My histories are always restored properly. Just wondering why this change would make this any better.

Maybe if you reorder panes, or delete some but the last one, or many other changes in the layout, you'll notice that your history is not so well restored.  
The `TMUX_PANE`/_pane_id_ instead is a unique identifier that is kept along the tmux session. Most of my changes target to keep them also across resurrections

Magnusgaertner:
> @javiplx interesting. I'm using the following with git master and it works splendidly:
> 
> ```
> [[ "$TMUX" ]] && TSWP=`tmux display -pt "${TMUX_PANE:?}" '#{session_name}:#{window_index}.#{pane_index}'` && HISTFILE=$HOME/.bash_history.d/bash_history-$TSWP
> PROMPT_COMMAND="history -a; $PROMPT_COMMAND"
> ```
> 
> My histories are always restored properly. Just wondering why this change would make this any better.

Thanks alot, but **make sure that ~/.bash_history.d/ exists** as a folder :)

PR #320, opened 2019-12-05
only add -l when default_shell is bash

tessus:
see https://github.com/tmux-plugins/tmux-resurrect/issues/94#issuecomment-562349123

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -100,7 +100,11 @@ tmux_socket() {
 # tmux server each time.
 cache_tmux_default_command() {
 	local default_shell="$(get_tmux_option "default-shell" "")"
-	export TMUX_DEFAULT_COMMAND="$(get_tmux_option "default-command" "-l $default_shell")"
+	local opt=""
+	if [ "`basename $default_shell`"x == "bash"x ]; then
+		opt="-l "
+	fi
+	export TMUX_DEFAULT_COMMAND="$(get_tmux_option "default-command" "$opt$default_shell")"
 }
 
 tmux_default_command() {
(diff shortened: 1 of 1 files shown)


scripts/restore.sh
```diff
@@ -100,7 +100,11 @@ tmux_socket() {
 # tmux server each time.
 cache_tmux_default_command() {
 	local default_shell="$(get_tmux_option "default-shell" "")"
-	export TMUX_DEFAULT_COMMAND="$(get_tmux_option "default-command" "-l $default_shell")"
+	local opt=""
+	if [ "`basename $default_shell`"x == "bash"x ]; then
```
bruno-: Can we use `$(basename $default_shell)` instead of backticks? Btw. what's `x` at the end of the strings used for?

tessus: > Can we use $(basename $default_shell) instead of backticks? 

Sure.

> Btw. what's x at the end of the strings used for?

It's an old habit to avoid test errors (unary operator issues), if one side is empty or undefined.
Theoretically `default_shell` should never be empty, also the double-quotes should prevent the error as well, but it all depends on the bash version and it doesn't hurt to be on the safe side.

tessus:
I suspect you squash the commits anyway. If not, I can create only one commit. Please let me know.

scripts/restore.sh
tessus: If you want I can remove the `x`.

bruno-:
This PR is merged via 1160c1da711c47f5b9a3983433442569f3c9d22b, I've removed the `x` suffix to strings.
Thanks

PR #328, opened 2020-02-13
Fix a race condition when restoring pane contents

jkp:
If the pane contents files are deleted too
quickly then pane contents restoration fails.

Cleanup previous content JiT before a restore
rather than after a restore has happened to
avoid this

### DIFF
--- a/scripts/helpers.sh
+++ b/scripts/helpers.sh
@@ -91,7 +91,12 @@ pane_contents_create_archive() {
 pane_content_files_restore_from_archive() {
 	local archive_file="$(pane_contents_archive_file)"
 	if [ -f "$archive_file" ]; then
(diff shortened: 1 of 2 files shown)


bruno-:
> If the pane contents files are deleted too quickly then pane contents restoration fails.

Can you please explain your understanding of how this happens?

 > Cleanup previous content JiT before a restore

How come there IS previous content? My understanding is we delete it after restore is completed.

jkp:
Hi - I cannot remember the details exactly now as it was a few weeks back but basically I found that if there is a lot of load on the system the contents files could be deleted before tmux got a chance to read from them. With this patch it never happens since old contents are deleted only when new ones need to be written.

Answering your second question: there can be previous content because with this patch it is deleted "lazily".

Issue #330, opened 2020-02-21
Pane titles are not restored


Avishayy:
Names that are set using `select-pane -T [name]` are not restored at all.

ioogithub:
I noticed that pane titles are not stored at all in the tmux_resurrect_dtm.txt files. If you work to set pane titles they will be lost after you resurrect.

Is this a bug or a missing feature?

PR #342, opened 2020-05-07
Restoring pane title

c3r34lk1ll3r:
Hi,

I made a little improvement adding the restoring of pane title. Simply, I added `#{pane_title}` to `pane_format` and `tmux select-pane -T <title>` in restoring.

This PR will fix also issue #330

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -169,7 +169,7 @@ new_pane() {
 
 restore_pane() {
 	local pane="$1"
-	while IFS=$d read line_type session_name window_number window_name window_active window_flags pane_index dir pane_active pane_command pane_full_command; do
+	while IFS=$d read line_type session_name window_number window_name window_active window_flags pane_index pane_title dir pane_active pane_command pane_full_command; do
 		dir="$(remove_first_char "$dir")"
 		window_name="$(remove_first_char "$window_name")"
 		pane_full_command="$(remove_first_char "$pane_full_command")"
@@ -194,6 +194,7 @@ restore_pane() {
 		else
 			new_session "$session_name" "$window_number" "$window_name" "$dir" "$pane_index"
 		fi
+    tmux select-pane -t "$session_name":"$window_number"."$pane_index" -T "${pane_title}"
 	done < <(echo "$pane")
 }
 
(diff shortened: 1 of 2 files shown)


ZachTheDev:
I would love to see this part of the next release!

neerajbadlani:
Tried this patch , named titles are saved and restored , howevers panes with empty titles are restored with current directory path . Can you please fix that too ?

scripts/restore.sh
```diff
@@ -194,6 +194,7 @@ restore_pane() {
 		else
 			new_session "$session_name" "$window_number" "$window_name" "$dir" "$pane_index"
 		fi
+    tmux select-pane -t "$session_name":"$window_number"."$pane_index" -T "${pane_title}"
```
bruno-, line 197: Instead of naming the pane with a standalone command, can we somehow set the correct pane_title when the pane is created?

scripts/save.sh
```diff
@@ -50,6 +50,8 @@ pane_format() {
 	format+="#{pane_pid}"
 	format+="${delimiter}"
 	format+="#{history_size}"
+	format+="${delimiter}"
+	format+="#{pane_title}"
```
bruno-, line 54: I'd prefer if we grouped all `pane_` interpolations together? Maybe add `pane_title` after `pane_index`?

This will be a breaking change when performing a restore with the old resurrect file, but that's a separate concern.

scripts/save.sh
```diff
@@ -313,7 +315,7 @@ main() {
 			start_spinner "Saving..." "Tmux environment saved!"
 		fi
 		save_all
-		if show_output; then
+    if show_output; then
```
bruno-, line 318: Can we please use tabs for indentation?

I'm also not happy with tabs and would like to change to two-space indentation, but I need help from someone that's willing to do that for all the repos in `tmux-plugins` github org.

bruno-:
Hi,

I'm in favor of adding this feature, but I don't have time to work on it.

I'll provide support and guidance if anyone is willing to step up and take the ownership of this feature. The code changes in this PR should already work so the next steps would be:

- Rebase with `master`
- Address PR comments
- Test, test, test
- Merge if all looks good

Hologos:
Hi @bruno-, I'll take care of this PR. Can I somehow continue in this PR or do I have to create a new one?

Hologos:
Just a note - for easier migration, I'll check number of saved pane parameters since you don't want the pane title to be last.

bruno-:
> I'll take care of this PR

Great!

> Can I somehow continue in this PR or do I have to create a new one?

Not sure, I think you have to create your own.

>  for easier migration, I'll check number of saved pane parameters since you don't want the pane title to be last.

We don't want to litter the plugin source code with migration logic. I'd prefer the `pane_title` parameter to be on line 51.

Hologos:
> We don't want to litter the plugin source code with migration logic.

That's a valid point, noted.

> I'd prefer the pane_title parameter to be on line 51.

Line 51 has delimiter on it, did you mean 52?

bruno-:
> Line 51 has delimiter on it, did you mean 52?

Ah, yes!

PR #361, opened 2020-09-18
Fix rename-window's target-window during restore

diegoximenes:
rename-window's target-window was not being properly set during restore, the session identifier was missing.
This was causing an unwanted naming pattern during restore.


bruno-:
Thanks

PR #362, opened 2020-10-02
Added support for saving and restoring named sessions as the default:

timmyjose:
1. The default session, "0" mimics the current behaviour, more or
     less.
  2. `C-l` in an active tmux session will list all the saved past
     "sessions" that you can choose to restore from/save to.
  3. If the target session is already running, simply attach to it, and
     save and restore as usual.
  4. If the target session is not running, simply rename the current
     session to the target session name, and then save and restore as
     usual.
  5. Alternately, simply start off with the session in mind:
     `tmux new -s <session-name>`, and then save and restore as
     usual.

### DIFF
--- a/resurrect.tmux
+++ b/resurrect.tmux
@@ -21,6 +21,14 @@ set_restore_bindings() {
 	done
 }
 
+set_list_resurrect_sessions_bindings() {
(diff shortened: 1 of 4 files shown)


timmyjose:
Wrong base.

Issue #250, opened 2018-05-21
“be careful” … of what?


isomorphisms:
`~/docs/restoring_programs.md` says:

> 
> - Restore **all** programs (be careful with this!):
> 
>         set -g @resurrect-processes ':all:'
> 

It would be helpful to be more specific about what can go wrong if one does this.

chrysn:
Concrete drive-by suggestions:

> Be *very careful* with this: tmux-resurrect can not know which programs take which context, and a `sudo mkfs.vfat /dev/sdb` that was just formatting an external USB stick could wipe your backup hard disk if that's what's attached after rebooting.

bruno-:
PR for docs improvement is welcome

PR #381, opened 2021-03-17
More explicit warning of restoring :all: programs

chrysn:
Closes: https://github.com/tmux-plugins/tmux-resurrect/issues/250

… adding the text suggested there and giving an example of why one *would* use that.


bruno-:
Thanks

PR #387, opened 2021-06-14
enable the tilde expansion ( aka ~ ) into $HOME in paths

YordanGeorgiev:
which is at least pretty value-able because of mac + vm with shared $HOME ... that is I could easily restore the same sessions between the shell in the host and the guest ...

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -130,6 +130,7 @@ new_window() {
 	local dir="$4"
 	local pane_index="$5"
 	local pane_id="${session_name}:${window_number}.${pane_index}"
+	dir="${dir/#\~/$HOME}"
 	if is_restoring_pane_contents && pane_contents_file_exists "$pane_id"; then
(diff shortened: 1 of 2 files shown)


scripts/restore.sh
```diff
@@ -130,6 +130,7 @@ new_window() {
 	local dir="$4"
 	local pane_index="$5"
 	local pane_id="${session_name}:${window_number}.${pane_index}"
+   dir="${dir/#\~/$HOME}"
```
bruno-, line 133: Can you use tabs for indentation here?

YordanGeorgiev: replaced with tabs, added .editorconfig to enforce "use tabs for spaces" rule ...

bruno-:
I merged this via 716b958145e2709e9fd3639ff24e1d1e4f26ff77

It contains the tilde expansion without .editorconfig.

Issue #389, opened 2021-07-08
Use non-global option @resurrect-dir instead of global option @resurrect-dir, this will allow maintaining multiple independent resurrect sessions


reportaman:
This will solve https://github.com/tmux-plugins/tmux-resurrect/issues/246.

Having ``@resurrect-dir`` to be local to a session:

```
tmux set @resurrect-dir '/Users/amanmehra/.resurrect/home/Desktop/textfiles/grid-typesetting'
instead of
tmux set -g @resurrect-dir '/Users/amanmehra/.resurrect/home/Desktop/textfiles/grid-typesetting'
```

 is the only missing link in maintaining and working simultaneously on different $PWD based resurrect sessions.

Once you switch over to using session local ``@resurrect-dir``, please also close https://github.com/tmux-plugins/tmux-resurrect/issues/246. A user can then write a 10ish lines bash script to be able to maintain multiple tmux-resurrect tmux sessions.

reportaman:
One option is to add a 3rd optional argument to get_tmux_option function in file [helpers.sh](https://github.com/tmux-plugins/tmux-resurrect/blob/master/scripts/helpers.sh):

```bash
get_tmux_option() {
	local option="$1"
	local default_value="$2"
    local showoptionswitch="${3:--gqv}"
	local option_value=$(tmux show-option $showoptionswitch "$option")
	if [ -z "$option_value" ]; then
		echo "$default_value"
	else
		echo "$option_value"
	fi
}
```

And a line in function resurrect_dir to:
```
		local path="$(get_tmux_option "$resurrect_dir_option" "$default_resurrect_dir" "-qv")"
```

This works! But with one caveat: tmux-resurrect saves all sessions. There should be an option to just store the current/focussed session in the "local" last session file.

reportaman:
PR has been created: https://github.com/tmux-plugins/tmux-resurrect/pull/390

bruno-:
As mentioned in #390 I encourage creating another plugin that focuses on saving/restoring individual sessions.

PR #390, opened 2021-07-08
Allow simultaneously working on and maintaining multiple resurrect sessions

### LINKED ISSUE #246: Is it possible to set `resurrect-dir` more dynamical
I am looking for a way to set `resurrect-dir` to the directory from which I start a new tmux session (extended with something like `.tmux_resurrect`).
I was hoping for a 'project based' resurrect workflow.

reportaman:
Changed helpers.sh and save.sh to allow simultaneously working on and maintaining multiple independent resurrect session files. Should resolve #246, and #389. Description of changes: 1) File helpers.sh: function get_tmux_option() now has an optional 3rd argument that can be used to pass custom option to tmux show-option command. This is used in function resurrect_dir() to pickup session local value of @resurrect-dir instead of using a global @resurrect-dir. Using session local values for @resurrect-dir is needed to maintain and simultaneously work with multiple different resurrect session files. 2) File save.sh: a missing filter has been applied to tmux list-sessions, tmux-listpanes, and tmux-listwindows. The filter does this: at time of saving a resurrect session, it checks if the current (attached) session is in a group. If that is true then the filter passes only sessions/panes/windows that belong to current (attached) group. Thus only current group's sessions/panes/windows are saved. If that is false, then filter passes only current (attached) session's panes and windows. Thus only current (attached) session's panes/windows are saved. Both changes to 1) and 2) together are required to enable working on multiple different resurrect sessions simultaneously, and independently. The effect of maintaining, working on multiple different resurrect sessions simultaneously can be achieved by the user by setting different values for @resurrect-dir. I confirm testing new capability.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ### master
+- Added support for simultaneously working on, and maintaining multiple independent resurrect sessions
+- Option @resurrect-dir has been changed from global to session local (use without -g)
 - save and restore tmux pane contents (@laomaiweng)
 - update tmux-test to solve issue with recursing git submodules in that project
 - set options quietly in `resurrect.tmux` script
--- a/docs/save_dir.md
+++ b/docs/save_dir.md
@@ -3,13 +3,13 @@
 By default Tmux environment is saved to a file in `~/.tmux/resurrect` dir.
 Change this with:
 
-    set -g @resurrect-dir '/some/path'
+    set @resurrect-dir '/some/path'
 
 Using environment variables or shell interpolation in this option is not
 allowed as the string is used literally. So the following won't do what is
 expected:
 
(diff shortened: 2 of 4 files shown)


reportaman:
In my testing things look perfect now, no case is missed, all possibilities have been tested. Now I would wait for your code review, and questions/suggestions.

bruno-:
Hey, thank you for the work on this PR.
This PR changes the plugin behavior a lot: instead of saving the whole environment (all the tmux sessions) it saves just a single session or a group of sessions.
As a current user of this plugin this breaks my workflow, so unfortunately these changes are unacceptable.

You now seem to understand well how tmux-resurrect works. I encourage you to fork the plugin (or just copy the code if you prefer that), rename it and focus on adding features relevant to saving/restoring individual sessions or projects. I would fully support this other projects by linking to it from tmux-resurrect docs.

Issue #343, opened 2020-05-13
Window names not changing after restore


Sviests:
Since I started using tmux-resurrect, I noticed that after restoring, the window names get stuck to what they were before the save and do not update afterwards, so I get left with generic window names like zsh and others, this became annoying once I started using fish shell, but the window name still shows zsh...

![image](https://user-images.githubusercontent.com/8798050/81781211-73967400-9500-11ea-8e65-cdf3109bd57b.png)

Thanks!

PR #399, opened 2021-08-21
Preserve the value of automatic-rename option

Farzat07:
When windows are renamed while restoring tmux, the value of
automatic-rename is set to off, which might not be desired.

Therefore, the original value of the automatic-rename option for each
window is now saved and set again while restoring. This means that
windows intended to have fixed names will continue to have fixed names,
while windows intended to have automatic-rename will continue to have
automatic-rename.

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -292,10 +292,11 @@ handle_session_0() {
 	fi
 }
 
-restore_pane_layout_for_each_window() {
+restore_pane_layout_and_autonaming_for_each_window() {
 	\grep '^window' $(last_resurrect_file) |
-		while IFS=$d read line_type session_name window_number window_active window_flags window_layout; do
+		while IFS=$d read line_type session_name window_number window_active window_flags window_layout automatic_rename; do
 			tmux select-layout -t "${session_name}:${window_number}" "$window_layout"
+			tmux set-option -t "${session_name}:${window_number}" automatic-rename "$automatic_rename"
 		done
 }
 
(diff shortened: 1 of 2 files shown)


Farzat07:
Fixes #57.
Fixes #343.

bruno-:
As mentioned in #222 we need a a right fix for this feature. Tmux's behavior around `automatic-rename` is unfortunately somewhat complex.

Farzat07:
Ok now I will tell you what I understand about the functionality and my plans for the new implementation, so please correct me if I am missing something.

The local auto-rename option can have one of three values:

- On, in which the window name is always auto-renamed.
- Off, in which the window name is always kept constant. Our problem is that renaming a window (which tmux-resurrect does) always sets this option to off, regardless of its initial value.
- Unset, in which case the option value is inherited from the global value.

The tmux format query #{?automatic-rename,on,off} returns only on or off. If the option is unset, the value of the global option is returned. This is the applies to both my implementation and The implementation in #222.

Ideally, one should keep the local option unset if that was initially the case. Actually that was my plan at the start, but I didn't implement it because I thought that the use case for this (changing the global option halfway through a session while the local options are unset) is not really meaningful (I can explain why but I guess this is not the place for that).

However, you are right. Even if I believe that this use case is not meaningful does not mean someone else won't find it useful. Handling that now would be better than doing it later.

Again if I am missing something, please let me know.

Farzat07:
I implemented the new ideas in #401.

Will close this PR.

PR #400, opened 2021-08-22
[WIP] Fix auto-rename after restore

smu160:
#222
#57

Hi @bruno-,

I was busy with work recently, but I should be able to knock out an initial solution in the next few days. I'm hoping to maintain this draft and collaborate on the fix until we get it right, so this PR is only draft for the time being. 

One of the first things I checked is how automatic-rename behaves given various settings. There are three potential outputs of `tmux show -w automatic-rename` that I noted:

1. Empty -- As you pointed out, this can be reproduced by first running `tmux set -u -w automatic-rename
`
2. `automatic-rename on` -- this can be reproduced by first running `tmux set -w automatic-rename on`
3. `automatic-rename off` -- this can be reproduced by first running `tmux set -w automatic-rename off`

Since we want to save the window-level values based on the result of  `tmux set -u -w automatic-rename`, I implemented a function following those three cases.

Please let me know if this is what we want. 

Thank you!

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -172,7 +172,7 @@ new_pane() {
 	else
 		tmux split-window -t "${session_name}:${window_number}" -c "$dir"
 	fi
-	tmux rename-window -t "${session_name}:${window_number}" "$window_name"
+	tmux rename-window -t "${session_name}:${window_number}"
 	# minimize window so more panes can fit
 	tmux resize-pane  -t "${session_name}:${window_number}" -U "999"
 }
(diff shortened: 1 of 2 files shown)


bruno-:
@smu160, thank you for starting the work on this. In the meantime @Farzat07 also started working on this in #401.

I don't know how to resolve this situation. Yes, you "signed up" for this work earlier, but then there was no communication for a couple weeks, and now Farzat started submitting PRs. I'll just let things happen and hope we have this feature implemented right and working soon 🤞

smu160:
@bruno-, this sounds great. I'll let @Farzat07 take the lead. I'll close out this PR and follow what is being done in #401. 

Thank you!

PR #401, opened 2021-08-23
Maintain the value of automatic-rename

### LINKED ISSUE #57: Window names are no longer dynamic after restore

### LINKED ISSUE #343: Window names not changing after restore

Farzat07:
An alternative implementation of #399.

Handles the unset case - if it was initially unset, it is kept that way.

Fixes #57.
Fixes #343.

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -286,6 +286,7 @@ handle_session_0() {
 }
 
 restore_window_properties() {
+	local window_name
 	\grep '^window' $(last_resurrect_file) |
 		while IFS=$d read line_type session_name window_number window_name window_active window_flags window_layout automatic_rename; do
 			window_name="$(remove_first_char "$window_name")"
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -292,10 +292,15 @@ handle_session_0() {
 	fi
 }
 
-restore_pane_layout_for_each_window() {
+restore_pane_layout_and_autonaming_for_each_window() {
 	\grep '^window' $(last_resurrect_file) |
-		while IFS=$d read line_type session_name window_number window_active window_flags window_layout; do
+		while IFS=$d read line_type session_name window_number window_active window_flags window_layout automatic_rename; do
 			tmux select-layout -t "${session_name}:${window_number}" "$window_layout"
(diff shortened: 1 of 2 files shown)


Farzat07:
Note that the -v option in "tmux show-window-options" removes the option name and only keeps its value (as desired).

scripts/restore.sh
```diff
@@ -292,10 +292,15 @@ handle_session_0() {
...
 }
 
-restore_pane_layout_for_each_window() {
+restore_pane_layout_and_autonaming_for_each_window() {
 	\grep '^window' $(last_resurrect_file) |
-		while IFS=$d read line_type session_name window_number window_active window_flags window_layout; do
+		while IFS=$d read line_type session_name window_number window_active window_flags window_layout automatic_rename; do
 			tmux select-layout -t "${session_name}:${window_number}" "$window_layout"
+			if [ "${automatic_rename}" = ":" ]; then
+				tmux set-option -u -t "${session_name}:${window_number}" automatic-rename
```
bruno-, line 300: I guess the only problem here is:

- We're first setting window name for each window in `restore_pane`

https://github.com/tmux-plugins/tmux-resurrect/blob/716b958145e2709e9fd3639ff24e1d1e4f26ff77/scripts/restore.sh#L180-L211

- Then `automatic-rename` is unset here (if needed).

That seems like duplicate work. As mentioned [here](https://github.com/tmux-plugins/tmux-resurrect/pull/222#issuecomment-892447307) I think we'll want to change where `window_name` is saved and restored.

bruno-:
@Farzat07 thank you for working on this 🙏

PR #422, opened 2021-12-19
Fix #141

### LINKED ISSUE #141: Error when restoring pane content

### LINKED ISSUE #192: cat: /Users/.tmux/resurrect/pane_contents//pane-jcs:5.1: No such file or directory

cartoonist:
As I mentioned in [#141](https://github.com/tmux-plugins/tmux-resurrect/issues/141#issuecomment-997377179), the issue apparently happens when using fish as the default shell. This commit fixes this issue by postponing `restore/pane_contents` clean-up after calling `restore_active_pane_for_each_window` (scripts/restore.sh:392). It might also fix #192.

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -270,9 +270,6 @@ restore_all_panes() {
 			restore_pane "$line"
 		fi
 	done < $(last_resurrect_file)
-	if is_restoring_pane_contents; then
-		rm "$(pane_contents_dir "restore")"/*
-	fi
 }
 
 handle_session_0() {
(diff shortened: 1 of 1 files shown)


bruno-:
Thanks for the investigation, testing and a fix.
Even though we're not sure *why* would this fix the problem, I'm ok merging it as it shouldn't change plugin functionality.

PR #443, opened 2022-07-15
Fix for restoring bash history

rbren:
This command seems to be more reliable - the other one I think grabs the "active" pane's title, which doesn't always sync up.


bruno-:
Thanks

PR #454, opened 2022-09-20
Add tmux note to keybinding for list-keys (now safer)

eggbean:
This PR simply adds a -N note so that the chosen key binding appears in `tmux list-keys` / `<prefix> + ?` and "Describe key binding" `<prefix> + /`, which is useful when you have a lot of custom keybindings.

### DIFF
--- a/resurrect.tmux
+++ b/resurrect.tmux
@@ -9,7 +9,8 @@ set_save_bindings() {
 	local key_bindings=$(get_tmux_option "$save_option" "$default_save_key")
 	local key
 	for key in $key_bindings; do
-		tmux bind-key -N "Save session" "$key" run-shell "$CURRENT_DIR/scripts/save.sh"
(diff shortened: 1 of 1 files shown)


sturob:
Using -N will fail to bind-keys on versions of tmux before 3.1 ([changelog](https://github.com/tmux/tmux/blob/149d9cc851dbf184b5b2041a5ec4be426b2f2914/CHANGES#LL604C1-L605C1)).

It is possible to [add the note after the binding](https://github.com/tmux/tmux/blob/149d9cc851dbf184b5b2041a5ec4be426b2f2914/CHANGES#L283). This would fail in tmux < 3.2, but the binding will still work.

eggbean:
I knew it failed with tmux < v3.1, but as that came out three years ago and there are newer versions even in the RHEL and Debian repositories, I thought it would be reasonable to add it, but I didn't know about being able to add the note without it causing errors.

I don't have any easy way to test it, but is this better now?

PR #455, opened 2022-09-20
ps.sh: fix ps arguments to work for busybox

tfaughnan:
Use the format specifier `'ppid,args'` instead of `'ppid command'`. POSIX allows either spaces or commas as separators, but busybox only allows commas. Furthermore, `args` is [recognized by POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/ps.html) while `command` is not. On implementations of ps that do recognize `command`, it is simply an alias for `args`, e.g. [Debian](https://manpages.debian.org/bullseye/procps/ps.1.en.html) and [FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=ps).

### DIFF
--- a/save_command_strategies/ps.sh
+++ b/save_command_strategies/ps.sh
@@ -11,7 +11,7 @@ exit_safely_if_empty_ppid() {
 }
 
 full_command() {
-	ps -ao "ppid command" |
+	ps -ao "ppid,args" |
 		sed "s/^ *//" |
 		grep "^${PANE_PID}" |
 		cut -d' ' -f2-


bruno-:
I think we should merge this change. The `ps` strategy is the default approach for saving commands.
Do you think this could break any of the existing systems?

tfaughnan:
I doubt it would break existing systems. I've tested (on an Arch Linux box) that it works for the ps provided by procps-ng, which is the default on many Linux distros. And the BSDs seem to agree with POSIX too judging by their man pages. Of course I could just install a non-busybox ps on my Alpine Linux system, but this change seems better to me.

PR #458, opened 2022-10-22
use XDG_DATA_HOME for resurrect-dir path

### LINKED ISSUE #348: Allow XDG standards on save-dir.
Pretty much the title. I just want to use `$XDG_DATA_HOME/tmux/resurrect` as my save directory.

Thanks for your hard work.

oliverlew:
To keep backwards compatibility, only consider XDG_DATA_HOME if `$HOME/.tmux/resurrect` is not found. Fixes #348

Not sure if that's all it needs. But anyway, I think this is an small change to benefit many XDG standard lovers, it should be supported sooner the better.


bruno-:
Thanks

PR #459, opened 2022-10-25
fix(#452): clear pane contents on restore

### LINKED ISSUE #452: Panes not "clean" on restore
So, tmux-resurrect works. I currently have ressurect set up and it correctly preserves pane contents. However, I notice that when I save a session and restore tmux, the panes are all "scrolled up."

As in, when I type several commands, clear the pane with <c-l>, save, and then restore, the pane's full history of those several commands is in view (rather than an empty pane as I saved it). Is there any way to alter this behavior?

barrett-ruth:
This seems to be a naive (but fully functional) approach to resolve #452.

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -202,6 +202,7 @@ restore_pane() {
 		fi
 		# set pane title
 		tmux select-pane -t "$session_name:$window_number.$pane_index" -T "$pane_title"
+		tmux send-keys -t "$session_name:$window_number.$pane_index" 'c-l'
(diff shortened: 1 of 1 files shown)


bruno-:
How do you know if a user's pane was cleared or not when they ran resurrect save?
Maybe there are users that don't like clearing panes.

barrett-ruth:
Similarly, though, there are certainly users who *did* clear panes and don't like to see their panes all polluted on restore.

I'll close this. Hopefully it helps anyone running into the same problem in the future.

PR #469, opened 2023-01-30
Restore with command arguments misc fixes

### LINKED ISSUE #468: Arrow (->) substitution preserving arguments
I use MacVim, so my vim processes are `/opt/homebrew/Cellar/macvim/.../MacOS/Vim`, which means that I need to use `~vim->vim` in my `ressurect-processes`. The problem with this is that the files I was editing don't get restored (the process ends up as just `vim`, when I'd want something like `vim main.go`).

If there was a way I could match and replace on _only_ the first argument (even regex would be fine), that'd be great!

ktprograms:
- Fix command being used as argument when there's no arguments (for example `vim` would turn into `vim vim`), due to the regex for removing the command name expecting there to be a trailing space (as there normally would be if there were arguments).
- Fix command and arguments being incorrectly parsed when the arguments contain a path, as the sed delimiter was `/`.
- Add docs on this restore option.

Fixes: #468

### DIFF
--- a/docs/restoring_programs.md
+++ b/docs/restoring_programs.md
@@ -28,6 +28,10 @@ contains space-separated list of additional programs to restore.
 
         set -g @resurrect-processes 'some_program "grunt->grunt development"'
 
+- Use `*` to expand the arguments from the saved command when restoring:
+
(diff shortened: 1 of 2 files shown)


bruno-:
Great PR, thank you!

PR #474, opened 2023-03-06
Update description of default session location

georgeanderson:
When explaining about session restore procedure, include all locations where the sessions can be stored. 

- Include alternate location `~/.local/share/tmux/resurrect/`
- Include reference to `${XDB_DATA_HOME}` env variable


bruno-:
Thanks

Issue #487, opened 2023-09-30
Tests failing on clean checkout from master.


mindhog:
I'm seeing test failures on a clean checkout from master (commit cff343cf9e81983d3da0c8562b01616f12e8d548).  To be fair, I can't entirely rule out my vagrant setup.  I had to disable UDP and switch to the the "peru/ubuntu-20.04-server-amd64" box, as the hashicorp box defined for ubuntu in tmux test doesn't work with the libvirt provider.

But the diff I'm seeing suggests a change in the file format:

```
< pane  0       0       :bash   1       :*      0       :/tmp   1       bash    :
< pane  blue    0       :vim    0       :       0       :/tmp   1       vim     :vim foo.txt
< pane  blue    1       :man    0       :-      0       :/tmp   0       bash    :
< pane  blue    1       :man    0       :-      1       :/usr/share/man 1       man     :man echo
< pane  blue    2       :bash   1       :*      0       :/tmp   1       bash    :
< pane  red     0       :bash   0       :       0       :/tmp   1       bash    :
< pane  red     1       :bash   0       :-Z     0       :/tmp   0       bash    :
< pane  red     1       :bash   0       :-Z     1       :/tmp   0       bash    :
< pane  red     1       :bash   0       :-Z     2       :/tmp   1       bash    :
< pane  red     2       :bash   1       :*      0       :/tmp   0       bash    :
< pane  red     2       :bash   1       :*      1       :/tmp   1       bash    :
< pane  yellow  0       :bash   1       :*      0       :/tmp/bar       1       bash    :
< window        0       0       1       :*      ce9e,200x49,0,0,1
< window        blue    0       0       :       ce9f,200x49,0,0,2
< window        blue    1       0       :-      178b,200x49,0,0{100x49,0,0,3,99x49,101,0,4}
< window        blue    2       1       :*      cea2,200x49,0,0,5
< window        red     0       0       :       cea3,200x49,0,0,6
< window        red     1       0       :-Z     135b,200x49,0,0[200x24,0,0,7,200x24,0,25{100x24,0,25,8,99x24,101,25,9}]
< window        red     2       1       :*      db81,200x49,0,0[200x24,0,0,10,200x24,0,25,11]
< window        yellow  0       1       :*      6781,200x49,0,0,12
---
> pane  0       0       1       :*      0       localhost       :/vagrant       0       bash    :
> pane  0       0       1       :*      1       localhost       :/tmp   1       bash    :
[...]

PR #488, opened 2023-10-04
Update tests for the latest code

mindhog:
Update tests to work with the latest versions of tmux-resurrect and tmux itself.


mindhog:
Fixes #487.  I can only test this under the `peru/ubuntu-20.04-server-amd64` vagrant box, since the existing Vagrantfile specifies box images that don't support libvirt -- can someone with the existing images run the tests and verify?

PR #547, opened 2025-07-27
Hotfix/keep 0 session alive

frogiraffe:


frogiraffe:
OOP SORRY MB. accident

PR #550, opened 2025-08-31
Create the default .tmux/resurrect dir if not exists

jmbeach:
Small change to create the resurrect folder if it doesn't exist.

Without this, the resurrect folder might get created in a folder the user doesn't expect. Ex: for me on my mac, the resurrect folder gets created at: `.local/share/tmux/resurrect`, but I thought it would be at `~/.tmux/resurrect`

I also got rid of the logic to fallback to `"${XDG_DATA_HOME:-$HOME/.local/share}"/tmux/resurrect` if `default_resurrect_dir` doesn't exist since it would make the logic a little more complicated to implement:

```diff
@@ -1,8 +1,4 @@
-if [ -d "$HOME/.tmux/resurrect" ]; then
-        default_resurrect_dir="$HOME/.tmux/resurrect"
-else
-        default_resurrect_dir="${XDG_DATA_HOME:-$HOME/.local/share}"/tmux/resurrect
-fi
+default_resurrect_dir="$HOME/.tmux/resurrect"
````

The problem is that the default folder only doesn't exist because it hasn't ever been created.

I figure this isn't really an issue though since users can already override the default by setting the `@resurrect-dir` setting.

### CODE REVIEW

File: scripts/helpers.sh [hunk 1]
```diff
@@ -98,9 +94,15 @@ pane_content_files_restore_from_archive() {
 
 resurrect_dir() {
 	if [ -z "$_RESURRECT_DIR" ]; then
-		local path="$(get_tmux_option "$resurrect_dir_option" "$default_resurrect_dir")"
+		local path
```
jmbeach, line 97: Fixes shellcheck lint: Declare and assign separately to avoid masking return values (SC2155)

jaseemabid:
Oh well I had the same issue so I updated the docs instead at #554 to describe the current behaviour.

PR #554, opened 2025-10-01
docs: Update default save directory

jaseemabid:
The default save directory changed from ~/.tmux/resurrect to
~/.local/share/tmux/resurrect in dd36a45, but the documentation was not
updated.

This commit updates docs/save_dir.md to document the current behavior
including XDG_DATA_HOME support and backward compatibility with the old
directory location.


jaseemabid:
cc: @xlucn

PR #557, opened 2026-01-29
Add per-session save files/session manager script

philippossfrn:
- Modified resurrect_file_path() to save sessions using session name
  - Prompts user to name session if using default numeric name
  - Added session manager script to $HOME/bin/ on install
    - Interactive menu to attach, delete, or rename saved sessions
  - Allows multiple distinct session saves instead of single global file

What the change does                                                                                                                                                                        
                                                                                                                                                                                              
  This modification enhances tmux-resurrect to save session state on a per-session basis rather than using a single global save file. Additionally, it includes a session manager script that 
  provides an interactive menu for managing saved sessions.                                                                                                                                   
                                                                                                                                                                                              
  Core changes:                                                                                                                                                                               
  - Modifies resurrect_file_path() to use the tmux session name as the save filename (e.g., ~/.tmux/resurrect/dev-project.txt instead of last)                                                
  - Prompts the user to name their session if it still has the default numeric name (e.g., 0, 1)                                                                                              
  - Installs a tmux-session-manager script to $HOME/bin/ that allows users to:                                                                                                                
    - Attach to a previously saved session                                                                                                                                                    
    - Delete saved session files                                                                                                                                                              
    - Rename saved sessions                                                                                                                                                                   
                                                                                                                                                                                              
  Why it's useful                                                                                                                                                                             
                                                                                                                                                                                              
  - Multiple project workflows: Keep separate save states for different projects without overwriting each other                                                                               
  - Cleaner organization: Session saves are named meaningfully instead of using timestamps                                                                                                    
  - Easy management: The session manager script provides a simple way to browse, clean up, or reorganize saved sessions without manually navigating the resurrect directory                   
  - Safer restores: Restoring one session won't accidentally load state from a completely different project                                                                                   
                                                                                                                                                                                              
  Breaking changes                                                                                                                                                                            
                                                                                                                                                                                              
  - Save file location changed: Existing saves in ~/.tmux/resurrect/last will not be automatically migrated. Users should manually restore their old session once, then save it under the new 
  naming scheme.                                                                                                                                                                              
  - Session naming prompt: Users with default-named sessions will now be prompted to provide a name on first save. This can be skipped by pre-naming sessions with tmux rename-session.

### DIFF
--- a/resurrect.tmux
+++ b/resurrect.tmux
@@ -31,10 +31,161 @@ set_script_path_options() {
 	tmux set-option -gq "$restore_path_option" "$CURRENT_DIR/scripts/restore.sh"
 }
 
+install_attach_script() {
+	local bin_dir="$HOME/bin"
+	local attach_script="$bin_dir/attach"
+
+	# Only create if attach script doesn't already exist
+	if [[ -f "$attach_script" ]]; then
+		return 0
+	fi
+
+	# Create $HOME/bin if it doesn't exist
+	mkdir -p "$bin_dir"
+
+	# Create the attach script
+	cat > "$attach_script" << ATTACH_SCRIPT
+#!/bin/bash
+
+# Directory where tmux-resurrect saves sessions
+RESURRECT_DIR="\$HOME/.tmux/resurrect"
+
+# Path to restore script (set at install time)
+RESTORE_SCRIPT="$CURRENT_DIR/scripts/restore.sh"
+
+# Check if the directory exists
+if [[ ! -d "\$RESURRECT_DIR" ]]; then
+    echo "Error: Tmux-resurrect directory not found at \$RESURRECT_DIR"
+    exit 1
+fi
+
+# List saved sessions
+echo "Available saved tmux sessions:"
+SAVED_SESSIONS=(\$(find "\$RESURRECT_DIR" -maxdepth 1 -type f ! -name "last" ! -name "readme*" -printf "%f\n" | sort))
+if [[ \${#SAVED_SESSIONS[@]} -eq 0 ]]; then
+    echo "No saved sessions found."
+    exit 0
+fi
+
+# Display session names
+for i in "\${!SAVED_SESSIONS[@]}"; do
+    echo "\$i) \${SAVED_SESSIONS[\$i]}"
+done
+
+echo ""
+echo "Options: [number] to restore, [d/D] to delete, [r/R] to rename"
+read -p "Enter your choice: " CHOICE
+
+# Handle delete option
+if [[ "\$CHOICE" =~ ^[dD]\$ ]]; then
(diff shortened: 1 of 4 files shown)


resurrect.tmux
```diff
@@ -31,10 +31,152 @@ set_script_path_options() {
 	tmux set-option -gq "$restore_path_option" "$CURRENT_DIR/scripts/restore.sh"
 }
 
+install_attach_script() {
+	local bin_dir="$HOME/bin"
+	local attach_script="$bin_dir/attach"
+
+	# Only create if attach script doesn't already exist
+	if [ -f "$attach_script" ]; then
```
laur89: Wouldn't mix posix `[]` with bash's `[[]]`. If we're writing bash, let's stay in bash syntax.
This applies not only to this line, but throughout this patch.

resurrect.tmux
```diff
@@ -31,10 +31,152 @@ set_script_path_options() {
...
+
+# Check if the directory exists
+if [ ! -d "$RESURRECT_DIR" ]; then
+    echo "Error: Tmux-resurrect directory not found at $RESURRECT_DIR"
+    exit 1
+fi
+
+# List saved sessions
+echo "Available saved tmux sessions:"
+SAVED_SESSIONS=($(ls "$RESURRECT_DIR" | grep -vE 'last|readme'))
```
laur89: `ls` is generally frowned upon in script usage. Best collect directory contents using `find`. That way there's also no need for the filtering `grep`, as find can provide that functionality.

resurrect.tmux
```diff
@@ -31,10 +31,152 @@ set_script_path_options() {
...
+# Check if 'last' is a symlink and remove it if necessary
+if [ -L "$RESURRECT_DIR/last" ]; then
+    rm "$RESURRECT_DIR/last"
+fi
+
+# Copy the selected session to 'last'
+cp "$RESURRECT_DIR/$SELECTED_SESSION" "$RESURRECT_DIR/last"
+
+# Ensure the tmux server is running by creating a detached session
+tmux new-session -d -s temp_session 2>/dev/null
```
laur89: Maybe I'm paranoid, but would randomize the temp session name just in case. E.g. `temp_session="temp-session-$RANDOM"`

philippossfrn: Good catch, also added a timestamp suffix to avoid creating more than one session with the same name

resurrect.tmux
philippossfrn: Nice thought, done.

resurrect.tmux
philippossfrn: Good point, fixed. Replaced all [ ] with [[ ]] for consistent bash syntax throughout the script.

PR #564, opened 2026-04-07
fix: prevent empty pane_title from corrupting save file fields

### LINKED ISSUE #437: Many bugs when restoring panes, pane placement and pane contents
I have been tracking several bugs over the course of the past month. I have saved and restored different environments over 100 times. There are several major bugs that I cannot find a way around. Some of the bugs happen every time. Some happen intermittently which suggests timing issue.

Several of the bugs may be related to adding new panes to window in that it messes up the existing panes, placements and histories. Sometimes adding a new pane to one window creates a new pane in a different window. Sometimes the panes are restored with a the placements messed up (almost always bunched together at the top).

Here is what I have found so far:

1. tmux-resurrect is creating phantom panes in the same window
- Create a new window with 3 vertical panes
- Save environment, kill server and restore
- 4 panes are restored to this window (an extra pane at the bottom)
- Sometimes it has the history of the 3rd pane and sometimes the history is empty.
- This is intermittent it happens around 20% of the time.

2. tmux-resurrect is creating phantom panes in different windows
- Create a new window with 3 vertical panes
- Save environment, kill server and restore
- The window is restored with 3 panes but an extra phantom pane is added to another window!
- This is intermittent, it happens around 15% of the time.

3. tmux-resurrect never restored the command to the bottom pane in a 3 window vertical split:
- Create a new window with 3 vertical panes
- Run the same script (~script.sh) to each window.
- Tmux will restore the script to pane 0 and 1, never 2
- Repeat the test with any script, it is always the bottom that is never restored regardless of the script type
- This is repeatable every time.

3b. tmux-resurrect never restored the command to the bottom pane in a 3 window vertical split:
- Swap pane 2 and 3
- Save environment, kill server and restore
- Observe that the third pane script is never restored, it doesn't matter what is run, it is the 3rd pane that is never restored.

4. tmux-resurrect sometimes remembers the commands of a 4 pane split window
- Create a new window with 4 vertical panes
- Each pane has the same script running (~script.sh)
- Save, kill server and restore
- tmux-resurrect restores the script to pane 1,2,4 (skips 3)
- repeat this test 5 times. It always skips pane 3.
- This is repeatable every time.

5. tmux-resurrect messes up the position of the windows
- In a window with 3 evenly spaced vertical panes, around 30% of the time the panes are restored all bunched up at the top.
- This may be related to adding a pane to another window.
- This is intermittent (happens around 20% of the time)

6. tmux-resurrect never remembers history of certain windows.
- The history of Window 0, pane 0 is never restored, all other windows and panes restore as expected.

There are so many different issues it is hard to track them all. I have never had a successful session completely restored, usually there are 3 or 4 issues that need to be corrected every time.

I would guess there are three main root causes:
1. A timing issue where sometimes the windows and panes are restore correctly and sometimes they are not. This might account for the intermittent nature of some of the issues.
2. Adding new panes seems to get tmux-resurrect confused, I think it often tries to restore the contents of the past pane position.
3. Even if no new panes are added, their are certain window configurations that are never restored properly (3 vertically split , evenly spaced panes) never restored the running command of the third pane.

KTSCode:
## Summary

When `pane_title` is empty (which tmux allows), `dump_panes()` in `scripts/save.sh` produces a save file line with consecutive tab delimiters. Bash's `read` with `IFS=$'\t'` collapses these consecutive tabs (POSIX whitespace-IFS behavior), shifting every field after `pane_title` one position left. This puts `pane_active` (0/1) into the `dir` field, and leaks `pane_pid` into `pane_command`.

On restore, the corrupted `dir` value defaults to `~/`, so panes end up in the wrong directory.

**Root cause proof** — bash collapses empty tab fields:

```bash
printf 'a\t\tb' | (IFS=$'\t' read -r x y z; echo "x=[$x] y=[$y] z=[$z]")
# x=[a] y=[b] z=[]  ← empty middle field lost, 'b' shifts left
```

## The fix

In `pane_format()`, change line 42 from:

```bash
format+="#{pane_title}"
```

to:

```bash
format+="#{?pane_title,#{pane_title},_}"
```

This tmux conditional format outputs `_` when `pane_title` is empty/unset, guaranteeing a non-empty field and preventing the IFS collapse. The `_` is harmless — untitled panes get title `_` on restore (they had no meaningful title to preserve).

## Why this approach

- **One file changed, one line changed** — minimal risk
- **Backward compatible** — existing save files with populated titles are unaffected; old save files without the `pane_title` field (pre-v4.0.0) still work since the field position hasn't changed
- **Non-destructive** — unlike prefixing a sentinel character that must be stripped on restore, this preserves all legitimate pane titles exactly as-is

## Related issues and PRs

- Fixes #437 — "When restoring, some panes of a session are corrupted"
- Alternative to #520 — that PR prefixes `:` to `pane_title` and strips all colons on restore, which is destructive to titles containing colons
- Alternative to #536 — that PR changes the delimiter from `\t` to `\t;\t` across multiple files, breaking backward compatibility with existing save files

## Test plan

- [x] Set a pane with an empty title (`tmux select-pane -T ""`) alongside panes with titles
- [x] Run tmux-resurrect save, inspect the save file — confirm no consecutive tabs in the `pane_title` field (empty titles saved as `_`)
- [x] Run tmux-resurrect restore — confirm all panes restore in the correct directory with correct commands
- [x] Verify panes with existing titles retain their titles after save/restore cycle


pdecat:
Hit this on a real snapshot, can confirm the diagnosis.

Env: tmux-resurrect at `cff343c` (current master), with tmux-continuum, 87 panes / 21 windows.

Pre-patch, the save file had 9 panes with exactly the field-shift signature described here — field 8 holding `0`/`1`, field 10 holding a real PID:

```
idx=2  path-field=[1]  cmd-field=[2425167]   <- bash pid
idx=1  path-field=[0]  cmd-field=[4187826]
idx=1  path-field=[0]  cmd-field=[9779]
... (6 more)
```

All 9 were panes where a TUI program had cleared `pane_title` between command exit and the next `PROMPT_COMMAND`, leaving the title empty at snapshot time. On restore the corrupted `dir` resolves to `$HOME`, which is the symptom that brought me here.

Applied the one-line patch locally, re-ran save: same 9 panes now serialize with title `_` and intact `pane_current_path` (paths match what adjacent panes had been hinting at). 0 misaligned entries in the new file.

PR #572, opened 2026-06-16
Ressurrect `copilot`, `agy` (antigravity) and `codex` CLI's 

viktorashi:
- Added this as an example of how to use hooks

You can add it via
```tmux.conf
set -g @resurrect-hook-post-save-layout '~/.config/tmux/scripts/resurrect-ai-cli-save.sh'
```
or whereever you want to put the script

This looks at the the respective apps' DB's and conversation state files and attempts to find exactly the conversation you were in before, and restores it via
```bash
agy --conversation=%s
```

```bash
copilot --resume=%s
```
and

```bash
codex resume %s
```
And it works really realiably and fast in my experience.

Yes, it's pretty long vibe-coded, but tbh it doesnt take a lot more time during save nor restore.

Potentially extensible to other AI TUI's.


viktorashi:
apparently #571  did the exact same thingg, LETS JOIN FORCES

PR #582, opened 2026-07-22
fix: empty pane title shifts every following field, restoring panes in the wrong directory

fdaciuk:
## The bug

Panes are restored in the default directory instead of the one they were saved in. It looks random, but it is fully deterministic: it happens to **every pane whose `pane_title` is empty** at save time.

## Cause

`pane_format()` emits `#{pane_title}` without the `:` sentinel that every other nullable field already carries:

```sh
format+="#{pane_title}"          # scripts/save.sh:42  <- no sentinel
format+="${delimiter}"
format+=":#{pane_current_path}"  # sentinel
```

Both `dump_panes()` (`save.sh:192`) and `restore_pane()` (`restore.sh:178`) parse those lines with `while IFS=$'\t' read ...`. Tab is an **IFS whitespace** character, so bash collapses runs of tabs into a single delimiter and empty fields vanish. Every field after the title shifts one position left:

```console
$ printf 'pane\tmain\t1\t0\t:-\t0\t\t:/Users/x/proj\t1\tzsh\t2533\t183\n' \
  | while IFS=$'\t' read lt sn wn wa wf pi pt dir pa pc pp hs; do
      echo "pane_title=[$pt]"; echo "dir=[$dir]"; echo "pane_active=[$pa]"
    done
pane_title=[:/Users/x/proj]
dir=[1]
pane_active=[zsh]
```

`dir` ends up holding `pane_active`. The saved line carries `dir` = `1`, and on restore `remove_first_char "1"` yields `""`. `tmux new-window -c ""` does not fail - it silently falls back to the default directory:

```console
$ tmux new-session -d -s t -c /Users/x/code; tmux new-window -d -t t -c ""
$ tmux list-panes -a -F '#{session_name}:#{window_index} [#{pane_current_path}]'
t:0 [/Users/x/code]
t:1 [/Users/x/dotfiles]   # cwd of the caller, not the saved path
```

Two more consequences:

- `restore_pane()` writes the shifted value back with `select-pane -T`, so panes end up literally titled `:/some/path`. That is corrupted data reinjected into the live server, and it persists across save/restore cycles.
- `dump_pane_contents()` (`save.sh:224`) reads the same format, so it gets an empty `history_size` and captures those panes with a bogus `-S` start line.

An empty `pane_title` is not exotic - it is what any program that resets the title on exit (`OSC 2` with an empty string) leaves behind:

```console
$ tmux send-keys -t t:1.0 "printf '\033]2;\033\\'" C-m
$ tmux list-panes -a -F '#{window_index} title=[#{pane_title}]'
0 title=[MacBookPro]
1 title=[]
```

## The fix

Prefix the field with `:` like the other nullable ones, and strip it back on restore. Two lines.

Verified against a live server holding a pane with an empty title - before the patch that pane saved `dir=[1]`, after it saves and restores the correct path, and the empty title round-trips faithfully.

## Compatibility note

Files saved before this change store the pane title without the sentinel, so the first restore from an old file drops the first character of each title. It corrects itself on the next save. Happy to guard it with `${pane_title#:}` instead if you would rather keep old files byte-compatible.

## Tests

I could not run the suite: `tests/helpers/helpers.sh` is a symlink into `lib/tmux-test`, an uninitialized submodule, and `tests/fixtures/*.txt` still describe a format that predates `pane_title` altogether. Both look pre-existing and out of scope here, but let me know if you want them refreshed in this PR.


fdaciuk:
Closing as a duplicate of #581.

I hit this independently and landed on the same two-line change before checking the open PRs. The resulting blobs are byte-identical - `save.sh` `01edcdeb..ad9d6ffd` and `restore.sh` `1a5e3f98..7fd48a2a` in both. #581 also refreshes the stale fixtures and adds regression coverage for the empty-title case, so it is strictly the better patch. Also related: #520, #536, #564, and the closed #570.

Leaving one data point here, since this bug now has several open PRs and no merge.

The empty pane title is not a corner case. In my setup it is produced routinely by Claude Code, which clears the title when it exits. Any program that emits `OSC 2` with an empty string leaves the pane that way:

```console
$ tmux send-keys -t t:1.0 "printf '\033]2;\033\\'" C-m
$ tmux list-panes -a -F '#{window_index} title=[#{pane_title}]'
0 title=[MacBookPro]
1 title=[]
```

There is also a second-order effect I did not see mentioned in the other PRs, and it is the reason this is hard to report cleanly: `restore_pane()` writes the shifted field back with `select-pane -T`, so an affected pane ends up literally titled `:/some/path`. That title is non-empty, so the *next* save parses correctly and faithfully records the directory the pane was wrongly restored into. The corruption disappears from the save file after one cycle while the pane stays in the wrong place, which makes it look intermittent and unreproducible. On a live server of mine, three panes had been carrying `:/Users/...` as their actual title for days.

Issue #563, opened 2026-02-23
Stacked pane layouts persist across save/restore cycles


clanghans:
<img width="905" height="129" alt="Image" src="https://github.com/user-attachments/assets/5681c5fd-2447-4298-850b-c188ec85f9fb" />
<img width="1262" height="174" alt="Image" src="https://github.com/user-attachments/assets/1f1e7c0e-1bcc-483c-ac90-30e72fe82202" />

## Problem

After restoring a session, multi-pane windows can end up with panes stacked — the first N-1 panes have a height of 1 row while the last pane takes all remaining space. Once this state is saved, it persists indefinitely across restore cycles.

## Observed Behavior

In affected save files, the layout strings for multi-pane windows encode pane heights of 1 for all panes except the last. This pattern matches what `new_pane()` produces with `resize-pane -U 999`, which temporarily minimizes panes during creation so more can fit. Under normal operation, `restore_window_properties()` applies the saved layout via `select-layout` to redistribute pane sizes after all panes are created.

In the observed case, the stacked layout was present in every available save file, including the oldest (suggesting it entered the save history early and was never corrected).

## Potential Contributing Factors

The following aspects of the current code could allow a stacked layout to enter the save file and persist:

1. **No layout validation after `select-layout`**: If `select-layout` fails or applies a layout that contains height-1 panes, nothing detects this. Panes remain stacked and the next save captures that state.

2. **`select-layout` errors are fully suppressed**: The call in `main()` uses `>/dev/null 2>&1` (line 372), making any failures invisible.

3. **No protection against saving during a restore**: There is nothing preventing tmux-continuum (or a manual save) from running while a restore is in progress. If a save occurs after panes are created but before `select-layout` completes, it would capture the intermediate stacked state.

4. **Self-perpetuating once triggered**: Once a stacked layout is saved, restoring it applies the broken layout via `select-layout` (which "succeeds" since the layout is technically valid), and subsequent saves capture the same state. There is no mechanism to break this cycle.

nullbyte61:
One concrete cause of this: **floating panes on tmux 3.6+**. With a floating pane open, `#{window_layout}` includes the floating pane twice, as an extra child of the root cell and in a trailing `<...>` section. tmux 3.7c rejects that string (`invalid layout`) even when fed straight back to the same window. The error is silenced in `restore_window_properties`, so every pane stays 1 row tall except the last, and the floating panes become tiled.

[...]

PR #593, opened 2026-09-15
Restore floating panes (tmux 3.6+)

nullbyte61:
tmux 3.6+ appends floating panes to `#{window_layout}` as extra children of the root cell, plus a trailing `<...>` section (omitted while a floating pane is zoomed). `select-layout` rejects that string, even fed straight back to the same window, and the error is hidden by `restore_window_properties >/dev/null 2>&1`. Windows with floating panes were restored with every pane stacked vertically, 1 row tall except the last, and the floating panes turned into tiled ones. That is one cause of #563.

**Fix (restore-side):** detect floating cells as the root children that follow once the tiled children fill the root cell. Skip those panes while creating panes, apply the layout without them (checksum recomputed), then recreate them with `new-pane -d -x W -y H -X x -Y y`. Floating panes always have the highest pane indexes in a window, so titles, pane contents and process restore still map to the right panes. Existing save files taken with floating panes restore correctly.

**Also includes #583** (commit by @josh-stephens, cherry-picked unchanged): empty pane titles collapsed the tab-separated fields, so untitled panes restored in the wrong directory. Floating panes set their title separately, so the follow-up commit strips the new `:` prefix there too. This commit changes the saved pane line. Titles in save files written before it lose their first character once on restore. If #583 is merged first, I'll rebase and drop it from here.

**Tested** on tmux 3.7c, save → kill server → restore on isolated servers, each case also run against `cff343c` to confirm it fails there:
- nested splits plus a float
- a single pane plus two floats
- an active float
- a zoomed float, and a zoomed tiled pane with floats present
- process restore inside floats
- running restore twice (no duplicate panes)
- untitled tiled and floating panes


nullbyte61:
Why #583 (@josh-stephens' commit) is part of this PR rather than left to its own:

Restoring a floating pane reads its title and working directory from the same saved `pane` line as every other pane. With an empty `#{pane_title}`, bash's `IFS=$'\t' read` collapses the empty field, and every later field shifts left. An untitled floating pane was recreated in the wrong directory, with its path applied as the title. The test for untitled tiled and floating panes fails on this branch without #583 and passes with it, so the floating-pane restore isn't correct without that fix.

The last commit here (af59689) depends on it too: it strips the `:` prefix that #583 adds, for floating panes, which set their title outside `restore_pane`. Without #583 that commit would drop the first character of every floating pane's title.

The #583 commit is cherry-picked unchanged, with its authorship kept. Only that commit was taken from #583's branch, not the two unrelated commits after it. If #583 is merged first, I'll rebase and drop it from here. If you'd rather review the two separately, #583 has to land before this PR.

nullbyte61:
Note on tmux 3.8: **this PR only matters for 3.7.x.**

tmux `bf43fdc0c7` (in the `3.8-rc` tag, not in 3.7c) replaces the custom layout format with a JSON one that includes floating panes, and `select-layout` accepts it. I built `3.8-rc` (`6f387ee`) and ran the same save → kill server → restore tests:

| tmux | resurrect | windows with floating panes |
| --- | --- | --- |
| 3.7c | master `cff343c` | restored stacked, floats become tiled |
| 3.7c | this branch | restored correctly |
| 3.8-rc | master `cff343c` | restored correctly |
| 3.8-rc | this branch | restored correctly |

On 3.8 the detection here finds no floating panes in a JSON layout and passes it through unchanged, so the branch is inert rather than harmful. It stays useful for anyone on 3.6/3.7, where `b368e89bea` ("disable custom layouts with floating panes for now") makes those layouts unrestorable.

One case still fails on 3.8-rc, identically with and without this branch, and it is a tmux bug rather than a resurrect one: with two or more floating panes, zooming a **tiled** pane sets every floating pane's z-index to the same value, and the duplicates persist after unzooming, so tmux rejects the window's own layout:

```
unzoomed z: "z":0 "z":1     select-layout: ok
zoomed z:   "z":1 "z":1     select-layout: duplicate pane z-index
unzoom z:   "z":1 "z":1     select-layout: duplicate pane z-index
```

Also reproduced on tmux master (`e880cf6`). Zooming a floating pane does not trigger it. I'll report it to tmux.

